PR #6, opened 2019-04-17
Blog: add Pygments highlighting and an Atom feed

rgommers:
Tested locally; feed will show up at http://labs.quansight.org/blog/feed.xml


rgommers:
Okay CI is triggered and output can be inspected. Still, can't see rendered site with TravisCI (see gh-6, we need Azure or CircleCI). Anyway, tested locally and we can see on TravisCI that the pages are generated. So merging.

PR #7, opened 2019-04-17
Set up CircleCI.

### LINKED ISSUE #4: Add a PR preview method
Otherwise things will break too easily.

GitHub Pages doesn't allow serving both from `master` and `gh-pages`, so perhaps easiest to switch to CircleCI which will store build artifacts and will render the whole site correctly in its UI.

hameerabbasi:
Fixes #4

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,11 +11,11 @@ jobs:
       - run:
           name: Install Lektor
           command: |
-            pip install Lektor
+            pip install --user Lektor
             lektor plugins add markdown-higlighter
             # lektor plugins add lektor-atom  # misses a bug fix, so (equal to
             # nixjdm/lektor-atom now, but pointing to a version we control):
-            pip install https://github.com/rgommers/lektor-atom/archive/master.zip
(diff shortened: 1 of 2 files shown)


hameerabbasi:
It seems CircleCI does build artifacts fine. 🙂 

However, it does not redirect/serve `directory` to `directory/index.html` automatically.

rgommers:
> However, it does not redirect/serve directory to directory/index.html automatically.

That's correct, you have to browse through Artifects and click on `index.html`. Then it server reliably though, and direct links to artifacts remain stable for quite a while.

This looks good. The navbar doesn't work, but I believe that's caused by having to use absolute rather than relative URLs for the `lektor-atom` plugin at the moment. Resolving that so won't worry about that here. Posts are looking fine: https://9-179573943-gh.circle-artifacts.com/0/home/circleci/project/_build/blog/2019/4/joining-labs/index.html

rgommers:
Merged, thanks @hameerabbasi

rgommers:
One issue with this, it triggers on all branches so failed in a chain reaction when the deploy to `gh-pages` triggered a next CI build. Need to specify to trigger on master and PRs only.

PR #8, opened 2019-04-18
Fix CircleCI deploy.

hameerabbasi:


hameerabbasi:
@rgommers Another reason to not use Lektor... The git deployment seems to be broken. I can push to the repo with my Personal Access Token from my local machine but with Lektor it hits a wall.

rgommers:
Yeah I wouldn't be surprised if it worked with TravisCI (because that's in the docs) but not out of the box with CircleCI.

Let's just merge this and make it work with Nikola (starting on that now).

PR #13, opened 2019-04-29
Add  metadsl blog post

saulshanabrook:
In this post I give some motivation for metadsl and build up a simple example from scratch. It doesn't use metadsl, but is the same concept so the hope is it can give readers a feel for it, without introducing too much complexity.

### DIFF
--- a/posts/2019/04/metadsl.ipynb
+++ b/posts/2019/04/metadsl.ipynb
@@ -1,0 +1,409 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# `metadsl`: A Framework for Domain Specific Languages in Python\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


rgommers:
Thanks Saul. Your post wasn't built because of the missing metadata. Could you add that (see https://github.com/Quansight-Labs/quansight-labs-site#creating-new-posts)?

hameerabbasi:
Actually, I made a mistake building only `master` even in PRs... 🤕 

It's fixed in https://github.com/Quansight-Labs/quansight-labs-site/pull/15/commits/f86dd5137afbbd34b301a10ac877248c8ace6c79

pearu:
Here's some feedback:

- typos: oeprations

- `Now we can start manually creating an expression graph` - why not just `expression` as `expression graph` is a visualization of an expression. If not, define graph.

- Use `def __add__(self, other: Number) -> Number:` instead of `def __add__(self, other: "Number") -> "Number":` to be less convolved.

hameerabbasi:
We should merge `master` into this because of CI fixes.

saulshanabrook:
> We should merge master into this because of CI fixes.

Will do.

saulshanabrook:
Thank you @rgommers  and @pearu for your comments on this. Hopefully it will build now. I fixed the weirdnesses you brought up and replaced all mentions of "graph" with "expression" to be more precise and consistent.

rgommers:
This looks good. Spotted one minor typo "consistancy". We'll wait a few days with publishing this one, to spread out the posts a bit (just did SymPy this morning, and uarray 2 days ago)

saulshanabrook:
We should hold off on publishing this until atleast after https://github.com/Quansight-Labs/metadsl/pull/57 is merged, so that the style in this blog post more closely matches the current implementation.

rgommers:
good timing, I had just put a reminder in my calendar about this post:) 

okay, please ping me when you want this published. also, before I forget again: remember to update the publication date, so it will actually be the first article people see and it gets on the Planets

rgommers:
Okay in it goes, thanks Saul!

PR #16, opened 2019-05-01
Fix CI.

hameerabbasi:


rgommers:
thanks! I guess that fixed the problem we saw on Aaron's PR?

hameerabbasi:
I certainly hope so, although someone without permissions may have to check.

rgommers:
let's see with the next PR then:)

hameerabbasi:
Or we can merge `master` into those other PRs. :)

rgommers:
Okay this is all good, deploy also worked as expected.

PR #34, opened 2019-08-19
Minor readme fixes

jrbourbeau:
This PR has a few very minor updates to the README:

- Adds quotes to `pip install nikola[extras]`, which can fail on some shells (e.g. zsh) that use square brackets for pattern matching

- Adds "https://" to labs.quansight.org, so the link is rendered

- Removes a few redundant ":"


rgommers:
thanks @jrbourbeau

PR #46, opened 2020-01-31
PR: Add non-blog config and some initial pages

dalthviz:
* Enables non-blog pages along with the blog posts. 
* Adds some example pages for the main page (`Home`) and about page (`About`).
* Point Blog section of the nav bar to the Blog index page instead of the Archive.

A preview:

web

PR #50, opened 2020-02-17
PR: Initial gallery support for pages

dalthviz:
Hi @rgommers and @koliphan2 , checking a little bit I found a plugin to embed galleries in `.rst` pages (like the ones I used to create the non-blog pages in #46). A preview:

gallery

I'm not totally sure if this is what we are looking for the use of galleries but let me know if this makes sens to you or if it could be useful :)


rgommers:
Thanks @dalthviz. Question: is it possible to add any text to this? I was thinking per gallery item something like:
- project name
- logo
- paragraph on how we support the project

If that's too much work, we can also just have the logos like this, and then in the rest of the page use standard Markdown with one section header per project + as much text as we think we need.

> checking a little bit I found a plugin to embed galleries in `.rst` pages

I guess that works for `.md` pages as well?

dalthviz:
@rgommers is possible to add text to the images by using the [`metadata.yml`](https://github.com/Quansight-Labs/quansight-labs-site/blob/05bc5b771bd5b409b9b47fce328599695c889630/galleries/projects/metadata.yml) and using the text in the [gallery template file.](https://github.com/Quansight-Labs/quansight-labs-site/blob/05bc5b771bd5b409b9b47fce328599695c889630/themes/quansightlabs/templates/gallery_directive.tmpl)

I think the plugin only works for `.rst` files but I could check :+1:

dalthviz:
@rgommers a new preview:

image

ccordoba12:
Could we add Spyder to the gallery? Just a suggestion.

koliphan2:
Would we want to use columns and boxes like how you've shown, or would having long rows be the way to go? 
Screen Shot 2020-02-21 at 4 52 25 PM

Screen Shot 2020-02-21 at 4 52 39 PM

Also, I could be mixing this up with something else, but did we want to list the specific contributors as well, or just what contributions were made?

troy-bailey:
I like the rows

dalthviz:
A new preview using rows (I think rows look better too):

image

rgommers:
Nice, +1 from me for the rows as well. That screenshot LGTM, gives some more space to add text about we're doing on the project. And visually a bit cleaner.

rgommers:
@dalthviz I'm happy with merging this as is, and then adding new content/project in a follow-up PR. Or keep it here, whichever is easier. Feel free to self-merge, I'm very much swamped this week due to the CZI workshop.

rgommers:
Here's a more complete list of projects we should put on. Projects we're actively working on:

Spyder
JupyterLab 
NumPy
JupyterHub
Numba 
XND
Dask
SymPy
PyData Sparse
Ibis
Uarray
Metadsl
Conda-forge
Altair
PyTables

And projects we aim to support (separate section header in between the lists above and below):

SciPy
Pandas 
Xarray
Matplotlib
Bokeh
Scikit-learn
Zarr

We may need to add a few more later

dalthviz:
@rgommers I think it could be good to merge this one. Maybe a new issue regarding the new content to be added and the distinction between already supported projects vs. aimed to support should be created; then I could start working on that too :+1: (just to keep matters of content vs layout/config isolated).

Also, I think I don't have write access to this repo so I can't merge this

dalthviz:
@koliphan2 maybe do you have write access to merge this one? if that is the case, could you merge it please? Thanks!

rgommers:
Great, thanks @dalthviz! I'll give you commit access as well.

PR #59, opened 2020-03-12
Move the blogposts into the right path by date.

hameerabbasi:
This PR moves a few blogs that were in the wrong directory into the right one.


hameerabbasi:
This might break a few links. @asmeurer Is it okay if it breaks the link to your blogpost?

rgommers:
> This might break a few links. @asmeurer Is it okay if it breaks the link to your blogpost?

Let's not do that, there are links floating around already, from internal Slack to Planet Python/SciPy, and maybe Twitter and other sources. it's not that important that the month numbering in the URL is off by one.

Thanks for noticing though Hameer

asmeurer:
Interesting. I guess this happened because the post was generated by nikola new_post, but then the date in the metadata was updated but not the path when we had the actual post date. It would be better if nikola generated the URL based on the metadata rather than the path for the md file. I wonder if there's a better workflow for drafting posts but then putting the correct date on them when the are published.

rgommers:
I did it by hand a few times rather than use `nikola new_post`, that doesn't matter much.

> It would be better if nikola generated the URL based on the metadata rather than the path for the md file

agreed. too lazy to open an issue though:)

> I wonder if there's a better workflow for drafting posts but then putting the correct date on them when the are published.

I typically push a tiny commit to update the date before hitting publish. I can just remember to move the directory if needed.

asmeurer:
It would be nice if there were a GitHub actions or something that put the date on the post when it goes to master. It's a problem I have with my own blog as well because I tend to start draft posts and then publish them many days later. 

This is not to say we need to build something like that here. It's more of a discussion for the nikola upstream.

PR #71, opened 2020-04-16
Add Google Analytics

### LINKED ISSUE #70: Add analytics
We should add some form on analytics, so that we can understand how different blog posts perform and if users are reading them. So it would be nice to know how long someone has each page open for.

The default option here is probably Google Analytics.

But we could opt for something open source, if we wanted... Then we might also have the ability to serve it on our own domain, and so not have it blocked by add blockers.

saulshanabrook:
Closes #70

### DIFF
--- a/templates/book.tmpl
+++ b/templates/book.tmpl
@@ -5,6 +5,16 @@
 <%inherit file="post.tmpl"/>
 
 <%block name="extra_head">
+    <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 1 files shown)


rgommers:
Thanks @saulshanabrook, merged.

rgommers:
@saulshanabrook I don't think this is working, as far as I can see it's saying zero visitors every day. Could you have a look?

saulshanabrook:
I think maybe I put it in the book template but not in the default one? I'll look into it.

PR #90, opened 2020-07-20
Blog Post: Designing with and for developers

isabela-pf:


rgommers:
Nice post, I like where this is going, thanks @isabela-pf.

I suspect quite a few readers may still wonder what that design process is actually like then. An example could help there. Not sure if you planned to add one here, or leave that to a follow up post (it could get large quickly with screenshots). I'm happy either way.

isabela-pf:
I've revised based on the above feedback and comments from @tonyfast. I reordered how everything is presented and made it clear that I won't be discussing design process much because when I started to write that it became longer and felt less focused to me (probably worthy of its own post if that is interesting).

rgommers:
> I've revised based on the above feedback and comments from @tonyfast. I reordered how everything is presented and made it clear that I won't be discussing design process much because when I started to write that it became longer and felt less focused to me (probably worthy of its own post if that is interesting).

Seems like a good decision to me. The separate post does sound quite interesting.

> Also, I can move this post to the appropriate August folder if needed. I didn't think of it ahead of time.

Thanks, that'd be helpful. Also, if you think it's ready you can edit the PR title to remove `WIP` (I often ignore WIP PRs if I'm busy).

isabela-pf:
I think I made all the appropriate changes. Thanks again for the review!

rgommers:
Looks great, thanks @isabela-pf! I reformatted the file, updated metadata, and merged. Should be live in a minute

PR #93, opened 2020-07-24
Add sponsor logos at the bottom of the front page

rgommers:
This is still incomplete, need to add:
- CZI
- Quansight (may be obvious, but maybe not - Quansight itself is still the largest sponsor)
- The D. E. Shaw group (after getting permission, via Anthony)


rgommers:
In it goes. Thanks for the help @steff456.

rgommers:
@steff456 how did you make the image? We'll need to edit it, we should add Quansight and CZI (first row). And in the future we probably need to add DE Shaw. If adding the source file (e.g. Inkscape) makes sense, let's do that. Otherwise, can you store it on Google Drive somewhere?

Also, if you could make the update for Quansight and CZI, that would be really helpful (next week is fine).

PR #95, opened 2020-07-25
Content for about page and front page

rgommers:
This could use some more content on the front page, and a graphic (e.g. like the world maps generated by HoloViz that we had on the doors for the retreat). I'll open separate issues for what's left.


rgommers:
Have to get this up (sponsor logos related), we'll review and improve later.

PR #104, opened 2020-07-27
PR: Update sponsors logos

steff456:
- [x] Add Quansight and CZI logos
- [x] Add some more space between logos.
- [x] Include the svg file in the images folder


rgommers:
Thanks @steff456. This is maybe overdoing it just a little, the graphic doubled in size:

image

Do you think you can try a 3 row layout? That would also shrink the logos in size, they're now very large.

steff456:
Do we have an exact desired height for the image?

rgommers:
I was hoping <= 1.5x larger vertically than the currently deployed version. So adding one row and a bit of whitespace. Adding horizontal and vertical whitespace together shouldn't increase the vertical size I'd expect.

PR #113, opened 2020-08-13
Added first draft of post about versioned-hdf5.

melissawm:


posts/2020/08/introducing-versioned-hdf5.md
```diff
@@ -0,0 +1,93 @@
...
+
+This file still doesn't have any data or versions stored in it. To create a new version, you can use a context manager:
+
+```py
+    >>> with versioned_file.stage_version('version1') as group:
+    ...     group['mydataset'] = np.ones(10000)
+```
+
+The context manager returns a h5py group object, which should be modified in-place to build the new version. When the context manager exits, the version will be written to the file. This creates a `_version_data` group containing metadata associated with versions in the file:
+
```
asmeurer, line 48: The `_version_data` group is an implementation detail. I would leave this out of the high level blog post. It should just mention here that you have to interact with the versioned groups and datasets using the versioned_hdf5 API, rather than trying to access the datasets using h5py.

posts/2020/08/introducing-versioned-hdf5.md
```diff
@@ -0,0 +1,93 @@
...
+    >>> v1['mydataset']
+    <InMemoryArrayDataset "mydataset": shape (10000,), type "<f8">
+```
+
+To access the actual data stored in version `version1`, we use the same syntax as h5py:
+
+```py
+    >>> dataset = v1['mydataset']
+    >>> dataset[()]
+    array([1., 1., 1., ..., 1., 1., 1.])
```
asmeurer, line 69: I would show an example of modifying `mydataset` in a second version and showing that you can access both versions of the dataset.

melissawm:
Thanks @asmeurer , those are great suggestions!

scopatz:
LGTM - @rgommers, how do you feel about it?

rgommers:
This reads really well, great post and excellent technical work @melissawm and @asmeurer!

I assume the use of the D. E. Shaw logo has been approved? If so, all that's left to do is add the `TEASER` comment (see repo README), update the date, and merge.

PR #115, opened 2020-08-21
Fixed authorship and added local image files.

melissawm:
As mentioned on #113.


rgommers:
Merged, thanks Melissa!

PR #121, opened 2020-08-31
Increase the body font-size to 18px

### LINKED ISSUE #109: Font size
The font size on the site feels rather small. I would suggest bumping it from the current 14px to at least 16px, if not more.

asmeurer:
Fixes #109.

### DIFF
--- a/themes/quansightlabs/assets/css/custom.css
+++ b/themes/quansightlabs/assets/css/custom.css
@@ -1223,7 +1223,7 @@ html {
 }
 body {
   font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
-  font-size: 14px;
+  font-size: 18px;
   line-height: 1.42857143;
   color: #333333;
   background-color: #fff;


rgommers:
Thanks Aaron. I agree that a font size bump would be good, but this seems a bit much. The code blocks in blog posts are overflowing (see, e.g., the most recent post on Traitlets).

How about trying 16px?

asmeurer:
16 seems fine. Although note that we can also easily adjust the font-size for code blocks separately.  

Also I noticed that the h4 font-size is already 18px so we should probably increase that (even with a body font-size of 16 it is too close IMO).

asmeurer:
These are the header font sizes

```css
h1,
.h1 {
  font-size: 36px;
}
h2,
.h2 {
  font-size: 30px;
}
h3,
.h3 {
  font-size: 24px;
}
h4,
.h4 {
  font-size: 18px;
}
h5,
.h5 {
  font-size: 14px;
}
h6,
.h6 {
  font-size: 12px;
}
```

So already h6 is smaller than body, which seems odd. This is what they look like on master

Screen Shot 2020-08-31 at 5 05 33 PM

rgommers:
I think no one should be using h5 or h6, and probably not h4 either. Such a deep nesting doesn't make sense for a normal page or blog post.

asmeurer:
I noticed the h4 thing because it's used in the Traitlets post.

rgommers:
Hmm yeah, I removed some section headers there, but it was pretty aggressive in using them.

rgommers:
@asmeurer want to rebase and update to 16px?

asmeurer:
Well, I'm not sure how to handle the headers. I think they would need to be increased too.

tonyfast:
Would it make more sense to REM units instead of fixed pixels?

asmeurer:
It would be nice if someone who actually understands design and/or CSS could finish this.

trallard:
Will close since this is already covered in #271

PR #122, opened 2020-08-31
Make Markdown the default for new posts

asmeurer:


rgommers:
Makes sense, thanks @asmeurer

PR #128, opened 2020-09-10
Fixed bad link on "Distributed and XLA device support"

vfdev-5:
Description:

- Fixed bad link on "Distributed and XLA device support" section


rgommers:
thanks!

PR #138, opened 2020-11-12
Tries to fix CircleCI build 

melissawm:
Changes pip upgrade strategy to solve nikola build problem on CircleCI.

edit: still trying to figure out the right solution to pass both CircleCI and Netlify.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/python:latest
+      - image: circleci/python:3.7
     steps:
       - checkout
       - run:


melissawm:
@rgommers Looks like pinning Python=3.7 to CircleCI works - this makes it compatible to the Netlify Python version.

rgommers:
Awesome, thanks Melissa!

rgommers:
That makes perfect sense - everything is pinned in `requirements.txt`, but can't pin the Python version there:(

rgommers:
There are two jobs; the deploy one on merge to master needed the same fix. Follow up in gh-139.

PR #140, opened 2020-11-13
Fix table and text.

ericdatakelly:
Here is a new version of @tonyfast's post on Ibis backends.  I used some html hacks and fixed up some other formatting.


rgommers:
Great, thanks @ericdatakelly! Merged.

PR #145, opened 2020-11-20
Adds link to full CZI proposal for NumPy, hosted on figshare.

melissawm:
Make sure we have a link to the full proposal to ensure transparency.


rgommers (approved):
Thanks Melissa!

PR #154, opened 2021-01-24
Blog post on Python packaging brainstorm and state of packaging today

rgommers:


posts/2021/01/packaging_painpoints_in_2021.md
```diff
@@ -0,0 +1,219 @@
...
+- Filling out configuration is tedious
+- Which freaking tool should I use!
+- Reproducible builds! (practically impossible) – IPython is reproducible for both `.tar.gz` anf `.zip`, but needs custom post-process.
+
+_RG: See [this blog post from Matthias Bussonnier](https://labs.quansight.org/blog/2020/08/ipython-reproducible-builds/) for more on this topic._
+
+- `setup.py` is dynamic making downstream packaging/automation much harder (including for conda) 
+    - Use [flit](http://flit.readthedocs.org/) – completely static! 
+
+_RG: For pure Python packages Flit is great indeed. For packages with compiled code, we'll be stuck with `setup.py` for a while longer I'm afraid. It is what it is - this has gotten better already with `pyproject.toml` and it's not one of the biggest issues anymore imho._
```
melissawm, line 78:
```suggestion
_RG: For pure Python packages Flit is great indeed. For packages with compiled code, we'll be stuck with `setup.py` for a while longer, I'm afraid. It is what it is - this has gotten better already with `pyproject.toml` and it's not one of the biggest issues anymore imho._
```

posts/2021/01/packaging_painpoints_in_2021.md
```diff
@@ -0,0 +1,219 @@
...
+- Shipping JupyterLab 3.0 extensions without Node.js is kinda cool.
+- Rust/Maturin to develop Python packages is nice.
+    - The relocation of depending binaries must be done manually in some cases.
+
+
+## topic 2: The conda world
+
+- Conda itself is almost unmaintained
+
+_RG: a quick peek at [the conda repo](https://github.com/conda/conda/pulse) indeed shows declining activity. Hopefully the new conda_
```
melissawm, line 97: I think this sentence needs to be completed.

posts/2021/01/packaging_painpoints_in_2021.md
```diff
@@ -0,0 +1,219 @@
...
+
+- The Conda and PyPA worlds still aren't talking to each other much.
+
+_RG: It's great to see that both the PyPA team and the Conda community are getting better organized. There's now more funding for work on PyPI, Warehouse and Pip; the teams working on it have grown, there's good project management, roadmapping and even Pip user experience research going on. And the Conda community has set up a new organization to align on standards, incubate promising new projects. The next step will hopefully be more productive interactions between those two somewhat separate groups._
+
+- `pip` and `conda` interaction still not great
+    - I'd like a way to have "safe" (e.g., pure Python) packages so you can confidently mix them. `+100`
+    - The pip dependency resolver is not great yet and it also ends up with conflicts with conda.
+
+_RG: True. For doing serious scientific computing, data science and ML type work, you often need the combination. Using `conda` for the core libraries, and filling in the gaps with `pip` for packages that are not packaged for conda or installing things from source._
```
melissawm, line 146:
```suggestion
_RG: True. For doing serious scientific computing, data science and ML type work, you often need the combination, using `conda` for the core libraries, and filling in the gaps with `pip` for packages that are not packaged for conda or installing things from source._
```

melissawm (commented):
Hi @rgommers - I made some small suggestions on things that Grammarly found for me :) Otherwise, my only comment is that as a reader I didn't find a good reference to read more on the native dependencies for wheels problem and how this is done currently. The same is true for the manylinux/Alpine Linux/musl comment. Can you link anywhere with a good write up of the current situation there? If not that's ok - I assume since this is a rather technical post, the people reading it will be familiar with the situation anyway.

mattip:
> The same is true for the manylinux/Alpine Linux/musl comment. Can you link anywhere with a good write up of the current situation there?

Since the perennial multilinux standard is based on the glibc version, and by design musl libc (which replaced glibc) does not have a version number, the Alpine Linux people must come up with a wheel versioning standard for a multibuild equivalent. For links, you could use [the realpython post](https://realpython.com/python-wheels/#the-manylinux-wheel-tag) which has a short blurb about musl/Alpine.

mattip:
>  I didn't find a good reference to read more on the native dependencies for wheels problem and how this is done currently.

This is a long-standing problem with shipping wheels. The naive answer is to ship any libraries your package needs in the wheel, which leads to multiple copies of so/dlls scattered in your site-packages. There are more subtle problems with this approach when two packages ship different versions of the same library. As far as I know the typical answer to this problem is "use conda" or "use the debian/homebrew/yum packages", which boil down to the same thing: find a single source of truth for packaging.

mattip:
Who is the target audience for this blog post? Maybe it should be divided into sections according to intended audience?

melissawm:
> Who is the target audience for this blog post? Maybe it should be divided into sections according to intended audience?

If the target audience is people who are already familiar with packaging and current problems around this in the Python world (as I suspect it is), maybe a sentence in the beginning making this explicit would be good.

rgommers:
> > Who is the target audience for this blog post? Maybe it should be divided into sections according to intended audience?
> 
> If the target audience is people who are already familiar with packaging and current problems around this in the Python world (as I suspect it is), maybe a sentence in the beginning making this explicit would be good.

Yes that's the target audience indeed. Good point, I'll make it clearer.

posts/2021/01/packaging_painpoints_in_2021.md
```diff
@@ -0,0 +1,219 @@
...
+packagers, library authors, and educators. This blog post contains the raw
+output of the 30 minute brainstorm (only cleaned up for textual issues), and
+my annotations on it (in italics) which capture some of the discussion during
+the session and links and context that may be helpful. I think it sketches a
+decent picture of the main pain points of Python packaging for users and
+developers interacting with the Python data and numerical computing ecosystem.
+
+<!-- TEASER_END -->
+
+## topic 1: The pip/wheel/PyPI world
```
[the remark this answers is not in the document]
rgommers: indeed - done

posts/2021/01/packaging_painpoints_in_2021.md
rgommers: thanks, fixed

rgommers:
> For links, you could use [the realpython post](https://realpython.com/python-wheels/#the-manylinux-wheel-tag) which has a short blurb about musl/Alpine.

That says no more than the one sentence I already had. I could find https://megamorf.gitlab.io/2020/05/06/why-it-s-better-not-to-use-alpine-linux-for-python-projects/ but that seems a bit negative. I'll leave it as is I think.

rgommers:
Okay let me get this up - my agenda is empty today, so it's a good day to post this content to the Twitterverse:)

rgommers:
Thanks for the reviews @melissawm and @mattip!

rgommers:
> Otherwise, my only comment is that as a reader I didn't find a good reference to read more on the native dependencies for wheels problem and how this is done currently.

This is worth another post at some point. At least I linked all of the tools involved here. Stuff like multibuild and delocate isn't well-known nor documented. https://cibuildwheel.readthedocs.io/en/stable/ seems better, but few projects in the PyData ecosystem use it.

rgommers:
Actually, that's not quite true anymore it seems, people are switching (e.g. scikit-learn, matplotlib, h5py): https://cibuildwheel.readthedocs.io/en/stable/working-examples/. Maybe we can nuke multibuild at some point .....

PR #167, opened 2021-03-11
Add the localsearch plugin

mattip:
as discussed on the qshare-tqz meeting this week. "European friendly" times make a difference!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,12 @@ auto: venv_nikola/bin/nikola ## build and serve the website, autoupdate on chang
 clean:  venv_nikola/bin/nikola  ## clean the website, usually not needed at all
 	venv_nikola/bin/nikola clean
 
+build_direct:  ## build using the local python3, for Netlify deployment
+	@python3 -m pip install wheel
+	@python3 -m pip install -r requirements.txt
+	@python3 -m nikola plugin -i localsearch
+	@python3 -m nikola build
+
 # Add help text after each target name starting with '\#\#'
 help:   ## Show this help.
 	@echo "\nHelp for building the website, based on nikola"
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,10 @@ all: build
 .PHONY: clean build help
 
 venv_nikola/bin/nikola:  ## create a virtualenv to build the website
-	@python3 -mvenv ./venv_nikola
-	@venv_nikola/bin/python -mpip install wheel
-	@venv_nikola/bin/python -mpip install -r requirements.txt
+    @python3 -m pip install --upgrade pip
+	@python3 -m venv ./venv_nikola
+	@venv_nikola/bin/python -m pip install wheel
+	@venv_nikola/bin/python -m pip install -r requirements.txt
 	@venv_nikola/bin/nikola plugin -i localsearch
 
 build: venv_nikola/bin/nikola  ## build the website if needed, the result is in ./public
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,29 @@
+SHELL := bash 
+TARGET ?= origin
+# pypy.org static page and blog makefile
+# type `make help` to see all options 
+
+all: build
+
+.PHONY: clean build help
+
+venv_nikola/bin/nikola:  ## create a virtualenv to build the website
+	@python3 -mvenv ./venv_nikola
+	@venv_nikola/bin/python -mpip install wheel
+	@venv_nikola/bin/python -mpip install -r requirements.txt
+	@venv_nikola/bin/nikola plugin -i localsearch
+
+build: venv_nikola/bin/nikola  ## build the website if needed, the result is in ./public
+	venv_nikola/bin/nikola build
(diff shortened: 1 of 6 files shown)


mattip:
The deployment fails. It needs to install the plugin. One way around this would be to add a Makefile so the CI and Netlify would do "make build" that also calls "make install" to install and build the site. The advantage is that then people could do "make build" to preview locally (if the "install" creates a virtual environment).

A simpler move would be to simply add `nikola plugin -i localsearch` to the netlify script :)

mattip:
Added a Makefile. @goanpeca I think is the admin of the netlify, so pinging for how to proceed. I guess one path forward would be
- change netlify to `make build` instead of `netlify build`, which will break everything except this PR
- make sure this PR successfully can deploy via Netlify
- review the search part of the PR
- merge it, which will unbreak everything

goanpeca:
@mattip see https://app.netlify.com/sites/quansight-labs/deploys/6058b074ac3deb2eef80a37c

goanpeca:
Screen Shot 2021-03-22 at 10 01 07 AM

mattip:
Maybe PyPI is having issues? Why would that fail? It comes from trying to create a virtualenv with
```
python3 -m venv ./venv_nikola
```

goanpeca:
I don't think we where using virtual envs there, '/opt/buildhome/python3.7/bin/python3.7 -m pip install --upgrade pip' 

So maybe we should just use whatever python is there?

goanpeca:
Agreed, but we should not need to have exactly the same setup for CI, local development, or netlify. Yes, having a single way of handling things would be nice, but I would not consider that a super strict requirement :) 

Screen Shot 2021-03-22 at 10 44 38 AM

mattip:
I added a `make build_direct` variant. Could you try that in Netlify?

goanpeca:
@mattip can you check all is working as expected https://deploy-preview-167--quansight-labs.netlify.app/ ?

Thanks :)

goanpeca (approved):
This is a great addition @mattip 

Thanks!

This LGTM 👍🏼

mattip:
It seems to work. The search bar in the right side of the navbar works as advertised and is fast. It searches locally, so it downloads 167.01 KB of text in the `tipuesearch_content.js` file, and then uses local CPU resources to actually scan and build the result. On https://www.pypy.org, with a couple of hundred blog posts, the download is 1.50 MB, and the search is still reasonably fast.

The search itself is exact, so searching for `Hameer` returns 3 results but searching for `Hammeer` does not return any.

goanpeca:
Unlese @rgommers has some input I would be inclined to merge this as is 🙃

rgommers (approved):
This looks and works great, thanks @mattip! And thanks @goanpeca for the review.

rgommers:
The deploy failed, one of the CircleCI jobs wasn't updated it looks like: https://app.circleci.com/pipelines/github/Quansight-Labs/quansight-labs-site/834/workflows/e8717154-a704-4ab2-bf75-37df4950f71e/jobs/833

Would you mind looking at this @mattip?

Makefile
```diff
@@ -0,0 +1,36 @@
...
+all: build
+
+.PHONY: clean build help
+
+venv_nikola/bin/nikola:  ## create a virtualenv to build the website
+	@python3 -m pip install --upgrade pip
+	@python3 -m venv ./venv_nikola
+	@venv_nikola/bin/python -m pip install wheel
+	@venv_nikola/bin/python -m pip install -r requirements.txt
+	@venv_nikola/bin/nikola plugin -i localsearch
```
asmeurer, line 15: I don't understand the purpose of having this here (or really of having the makefile at all). This command just creates plugins/localsearch. Why not just commit that to the repo?

asmeurer:
This whole things seems to be adding unnecessary complexity to the site build process. What's wrong with just using the niikola commands directly? The make commands all seem to just call out to nikola, so what's the point?

mattip:
I would think in the future we might go beyond the borders of Nikola and have some of our own styling. For instance, in the www.pypy.org/blog I use the [sidebar plugin](https://github.com/getnikola/plugins/tree/master/v7/sidebar) to provide a side navbar. This needs a bit of post-processing to update the static site (it could be done server-side or in javascript too) which necessitated an extra build step outside Nikola. This also brings the workflow for netlify preview deploys into the control of this repo (via `make build_direct`), otherwise if something changes in the netlify workflow in a PR then you need to coordinate with the netlify owner.

But whatever makes it easier for people to use and understand. I think you are correct that commiting [the plugin](https://github.com/getnikola/plugins/blob/master/v7/localsearch) to the repo would be sufficient for now.

Makefile
rgommers: I just ran into the same thing. Sorry, I was in a hurry and didn't review this PR closely enough. Invoking `pip install` and downloading packages on every invocation of `make auto` is a big no-no. I'll open a separate issue for this.

PR #168, opened 2021-03-11
Add ndindex to the list of projects

asmeurer:
Not sure if there's any order for the projects, so I just added it to the end.


rgommers (approved):
LGTM, thanks @asmeurer. Putting it at the end seems good.

PR #172, opened 2021-03-23
fix circleci deploy step

mattip:
@rgommers

PR #173, opened 2021-03-23
activate the venv since 'deploy' requires PATH modification

mattip:
Redo #172 to activate the venv instead


rgommers (approved):
Thanks Matti, let's try this once more:)

mattip:
Success!

PR #179, opened 2021-04-09
Fix example 5 code

kurtamohler:
Of course, the day the post goes live, one of the examples breaks. `make_reduction` no longer accepts undefined result tensors. This change is shown to fix the issue here: https://github.com/kurtamohler/pytorch-TensorIterator-examples/actions

cc @rgommers


rgommers (approved):
Thanks for noticing and the quick fix Kurt!

PR #181, opened 2021-04-11
Assume 'make build' is run after 'pip install -r requirements.txt'

### LINKED ISSUE #178: remove Makefile and handle localsearch dependency better
See https://github.com/Quansight-Labs/quansight-labs-site/pull/167/files#r610653619 for details. @mattip would you be able to take care of this?

mattip:
Also be more cache friendly for circleCI.

Closes #178

I left the Makefile in order to abstract the netlify preview deploy: it still can use "make build_direct".


mattip:
@rgommers

rgommers:
> I left the Makefile in order to abstract the netlify preview deploy: it still can use "make build_direct".

I'm not sure I understand, there is no usage of `make` anywhere as far as I can tell, and `make build_direct` is just an odd alias for `nikola build`?

mattip:
`make build_direct` is used by the netlify deploy. You don't see it because it is part of the netlify UI. It seems like just using `nikola build` as a netlify ocmmand is too limiting, I think we want to have a generic command there that we control via this repo, so leaving it as `make build_direct` (or changing it to `make netlify_preview` to be even more specific) makes sense to me, but whatever makes sense to you. If you want to change it, @goanpeca is the admin for the netlify build.

rgommers:
Ah fun. Would be a lot nicer if Netlify just had a config file that we could commit.

Okay, let's leave as is for now then, and merge this. Thanks Matti.

PR #183, opened 2021-04-11
Fix file name and slug, should be lowercase for deployment

rgommers:
Let's see if that makes this blog post show up.


ccordoba12 (approved):
Thanks @rgommers! That made the trick (at least on Netlify).

rgommers:
Netlify was working before, not sure what the problem is. Having `build` and `build_deploy` diverge is quite annoying.

ccordoba12:
> Netlify was working before

I think it wasn't, at least that's what I saw last night. But I didn't report anything because I thought something other was required behind the scenes during deployment (will report for next time).

> Having build and build_deploy diverge is quite annoying.

Agreed.

rgommers:
This worked: https://labs.quansight.org/blog/2021/04/a-step-towards-educating-with-spyder/

rgommers:
> I think it wasn't, at least that's what I saw last night. But I didn't report anything because I thought something other was required behind the scenes during deployment (will report for next time

I know it wasn't last night, but that was because the publication date was set to today - if the date/time is in the future, the post won't show up. On the PR it did show up in preview today.

ccordoba12:
Ok, I see. Then the problem is really having two different build configurations.

PR #188, opened 2021-04-18
replace circleci with github actions for deploy

mattip:
Fixes #187 

Since this destructively replaces the gh-pages branch and a mistake will ruin the website, please carefully monitor the site while/after merging this. Probably best done together with publishing a new blog post.

Also does not cache after `pip install`, maybe that should be left for a follow on PR to minimize chances of doing damage.

PR #194, opened 2021-05-07
Build website using github action.

Carreau:
Redo of #188 that had a few shortcoming and limitation (like nuke history, and remove CNAME and other critical files). 

Triggers only on master; (and hence why it's a PR from _my_ master branch to this master branch, so that one can go to [here](https://github.com/Carreau/quansight-labs-site/actions) to check that it did run on my fork, and that it pushed a single commit to my gh-pages branch (https://github.com/Carreau/quansight-labs-site/tree/gh-pages), and was therefore also deployed as a fork there: https://carreau.github.io/quansight-labs-site/blog/, so that it's really likely to work once merged.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,32 @@
...
+    - name: Deploy
+      if: github.ref == 'refs/heads/master'
+      shell: bash
+      run: |
+          set -eou pipefail
+          git config --global user.name 'GithubAction'
+          git config --global user.email 'githubaction@invalid'
+          git pull --all
+          git remote add deploy https://${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY.git
+          nikola github_deploy -m 'Nikola auto deploy [ci skip]'
```
goanpeca, line 30: How about using https://github.com/marketplace/actions/deploy-nikola-to-github-pages

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,32 @@
...
+      run: nikola build
+    - name: Deploy
+      if: github.ref == 'refs/heads/master'
+      shell: bash
+      run: |
+          set -eou pipefail
+          git config --global user.name 'GithubAction'
+          git config --global user.email 'githubaction@invalid'
+          git pull --all
+          git remote add deploy https://${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY.git
```
rgommers, line 29: A token shouldn't be needed within the same repo, not having to deal with authentication is one of the main advantages of using GH Actions right? Also, the secrets don't actually exist on this repo?

Carreau: It was my understanding that `secrets.GITHUB_TOKEN` is a magic token that automatically exist for workflow running on pushed branches (and not pull-requests), but let's just try.
```suggestion
          git remote add deploy https://github.com/$GITHUB_REPOSITORY.git
```

Carreau: > It was my understanding that `secrets.GITHUB_TOKEN` is a magic token that automatically exist for workflow running on pushed branches (and not pull-requests), but let's just try.

You were right, it works without.

rgommers:
Let's give it a try! Thanks @Carreau

rgommers:
Let's leave the CircleCI trigger enabled until we see this working.

PR #196, opened 2021-05-11
Rebuild the website daily.

Carreau:
From the documentation this should trigger on the last commit of the
default branch.

This will allow to schedule posts and have them auto published.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - master
+  schedule:
+      # * 6:37 every day – 37 just to not trigger exactly at 6:30 like many other;
+      # 6 is a bit before the beginning of the day in the US.
+      - cron:  '6 37 * * *'
 
 jobs:
   nikola_build:
@@ -27,5 +31,5 @@ jobs:
           git config --global user.email 'githubaction@invalid'
           git pull --all
           git remote add deploy https://github.com/$GITHUB_REPOSITORY.git
-          nikola github_deploy -m 'Nikola auto deploy [ci skip]'
+          nikola github_deploy -m 'Nikola github auto deploy [ci skip]'
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -5,9 +5,9 @@ on:
     branches:
       - master
   schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - master
+  schedule:
+      # * 6:37 every day – 37 just to not trigger exactly at 6:30 like many other;
+      # 6 is a bit before the beginning of the day in the US.
+      - cron:  '6 37 * * *'
```
rgommers, line 10: This looks like it's the wrong way around: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule

https://crontab.guru/#6_37_*_*_*

Carreau:
```suggestion
      - cron:  '37 6  * * *' 
```

Darn, cron should take named arguments.

rgommers: Comment is also still wrong I think? First link about says the time is in UTC, which means it'll be around midnight in US time zones.

.github/workflows/deploy.yml
```diff
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - master
+  schedule:
+      # * 6:37 every day – 37 just to not trigger exactly at 6:30 like many other;
+      # 6 is a bit before the beginning of the day in the US.
+      - cron:  '37 6  * * *'
```
Carreau, line 10:
```suggestion
  schedule:
      # * 10:37 every day – 37 just to not trigger exactly at 10:30 like many other;
      #  10:37 UTC is 6:30am EST is a bit before the beginning of the day in the US,
      - cron:  '37 10  * * *' 
```

.github/workflows/deploy.yml
Carreau: Ok, move to 10:37 - 6:37EST – 3:30PST. I think that does not matter much as long as it's at night.
10:30  is still early enough for Europe I think, and we can still tweak.

rgommers (approved):
Let's give this a try:) Thanks Matthias!

rgommers:
Weirdly this worked for a few days and then failed, without showing any CI run log:

image

No commits were made to master in the meantime.

PR #199, opened 2021-05-26
STY Makes the header consistent between posts and other pages

thomasjpfan:
Makes the header consistent between posts and other pages

On `master`:

Screen Shot 2021-05-26 at 12 28 57 PM

This PR:

Screen Shot 2021-05-26 at 12 29 10 PM


rgommers (approved):
Thanks for tracking that down Thomas!

PR #212, opened 2021-06-22
[July 7] pyflyby blog post

Carreau:


Carreau:
We still need to converge on a date and time.

Carreau:
This is now setup to be autopublished on Friday, July 2cd. 

The final URL should be https://labs.quansight.org/blog/2021/08/better-interactive-jupyter-sessions-with-pyflyby/

Carreau:
Hum I think we should move it to Fri 9 or Tue 13th, as this week-end is independece days week end, so people are likely to miss it as they will be away or catching up with monday off.

Carreau:
We talked at the content creator meeting, and Wednesday 7  would be a good day as well, but not the 6 as 5th is holidays in US.

posts/2021/08/better-interractive-jupyter-session-with-pyflyby.md
```diff
@@ -0,0 +1,202 @@
...
+```bash
+$ find-import norm
+from scipy.stats.distributions import norm
+```
+
+# Notes on pyflyby codebase
+
+The [Pyflyby codebase](https://github.com/deshaw/pyflyby) provides a window into
+advanced data structures and programming paradigms. Its use of modules and
+programming concepts are unusual relative to those found in more classical data
```
juanis2112, line 152:
```suggestion
programming concepts is unusual relative to those found in more classical data
```

posts/2021/08/better-interractive-jupyter-session-with-pyflyby.md
```diff
@@ -0,0 +1,202 @@
...
+ - Pyflyby demonstrates the use of  [Aspect-Oriented
+   programming](https://en.wikipedia.org/wiki/Aspect-oriented_programming),
+   which highlights the flexibility of the Python programming model.
+
+
+# Conclusion
+
+Pyflyby's utilities are designed to improve developer efficiency and materially
+reduce the impact of interruption on productivity. Its value, however, is
+broader than that. Pyflyby expands what is capable within the Python ecosystem
```
juanis2112, line 169:
```suggestion
broader than that. Pyflyby expands what one is capable of doing within the Python ecosystem
```
"what is capable" sounds a bit weird specially because it is missing a subject so I suggest "what one is capable of doing" instead but maybe something like "what we can do within the Python ecosystem...." could work too.

juanis2112 (commented):
I think this is ready minus some suggestions and one typo

posts/2021/08/better-interractive-jupyter-session-with-pyflyby.md
Carreau: That also looked weird to me, but as this was in a change D.E. Shaw suggested so I left it originally.

Will commit all your suggestions.

PR #218, opened 2021-07-07
Allow to manually trigger rebuild

Carreau:
Merging this will also release the Pyflyby blog post that _should_ already have been published.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,6 +1,7 @@
 name: Deploy built master on gh-pages
 
 on:
+  workflow_dispatch:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


rgommers (approved):
Not 100% sure from the docs if this works (`workflow_dispatch:` always has an indented block in the docs I skimmed). But easiest way to learn is to give this a try.

Thanks @Carreau

rgommers:
Guess it works:

image

Carreau:
Thanks !, Yes, I was not sure either, but I assume the PR would fail if that was invalid.

rgommers:
_(close eyes and deploy to prod)_

PR #223, opened 2021-07-13
Enable utterances (comment system) on the blog.

Carreau:
Before merging:
 - [x] A label named "utterances" need to be created in the repo
 - [x] The utterance bot needs to be installed on this repo via https://github.com/apps/utterances/installations/new

-- 

[Here](https://elegant-bhaskara-0c18b3.netlify.app/blog/2021/07/pyflyby-improving-efficiency-of-jupyter-interactive-sessions/) is how it looks like (on a fork of mine, with a different config to comment on a different repo), and [here is the issue](https://github.com/Carreau/quansight-labs-site/issues/2) the utterance bot create on the fork.

(You can try to comment on above demo deployment, but be carefull the redirect is incorrect, and you will need to mess with the URL once you've authorized github)

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -896,12 +896,12 @@ RSS_COPYRIGHT_FORMATS = CONTENT_FOOTER_FORMATS
 # systems.  The following comment systems are supported by Nikola:
 #   disqus, facebook, intensedebate, isso, muut, commento
 # You can leave this option blank to disable comments.
-COMMENT_SYSTEM = ""
+COMMENT_SYSTEM = "utterances"
 # And you also need to add your COMMENT_SYSTEM_ID which
 # depends on what comment system you use. The default is
 # "nikolademo" which is a test account for Disqus. More information
 # is in the manual.
-COMMENT_SYSTEM_ID = ""
+COMMENT_SYSTEM_ID = "Quansight-Labs/quansight-labs-site"
 
 # Create index.html for page folders?
 # WARNING: if a page would conflict with the index file (usually
@@ -1278,7 +1278,15 @@ WARN_ABOUT_TAG_METADATA = False
 
 # Put in global_context things you want available on all your templates.
 # It can be anything, data, functions, modules, etc.
-GLOBAL_CONTEXT = {}
+
+GLOBAL_CONTEXT = {
+    "utterances_config": {
+        "issue-term": "title",
+        "label": "utterances",
+        "theme": "github-light",
+        "crossorigin": "anonymous",
+    }
+}
 
 # Add functions here and they will be called with template
 # GLOBAL_CONTEXT as parameter when the template is about to be


mattip:
I like utterances. It will add an issue to the repo and then the discussion of the issue is rendered as comments on the blog post. That means all the authentication is done via github. The issues are created only as needed (when the first comment is made), but it does clutter up the repo a bit.

Carreau:
> I like utterances. It will add an issue to the repo and then the discussion of the issue is rendered as comments on the blog post. That means all the authentication is done via github. The issues are created only as needed (when the first comment is made), but it does clutter up the repo a bit.

We can make a separate repo for the comments, like we could have "Quansight-Labs/blog-comments" to not clutter the repo.

rgommers:
I like the look of this, and that it uses GitHub auth (prevents spam).

The "Act on your behalf" permission is a little vague, I wouldn't click on that without investigating first. I'm guessing it just wants to make commits with my author name, but it's unusually non-specific for a GitHub permissions request

image

mattip:
One option to mitigate possibly security concerns would be to make a dummy github robots.quansight@quansight.com account  and use it rather than @rgommers to register utterances with this repo.

rgommers:
That doesn't quite make sense, I'm just trying to make a comment. My screenshot is what every reader who wants to comment will see.

mattip:
ahh. less good.

Carreau:
"Act on your behalf" in this particular circumstance is:

 -  Read access to metadata
 -  Read and write access to issues

You can look at https://github.com/Carreau/quansight-labs-site/issues/2 and see what it looks like:

Screen Shot 2021-07-14 at 11 55 53

Utterance being a GitHub App, its permission are only the ones that each repo maintainer give it when installing the bot on the repo which I did on my fork:

Screen Shot 2021-07-14 at 11 56 58

Notice how the comment say that it has been done on my behalf and not directly.

Yes the GH prompt is confusing, I agree, but the permissions are extremely restrictive, the prompt could be better.
We can (and should) also link directly to the issue, in which case user can comment directly on the given issue, w/o going through utterances – one caveat though is that the utterance bot create the issue on first comment I believe.

mattip:
> one caveat though is that the utterance bot create the issue on first comment

We could, as part of the deploy script, create issues for all posts

Carreau:
> We could, as part of the deploy script, create issues for all posts

Yeah, there always are workaround for sure; by "issue" I only meant something which is currently not automatic/done.

rgommers (approved):
I have created the `utterances` label and installed the app. Happy for this to be merged so we can start using it. Would be nice to make the user experience as friendly as possible with one of the suggested approaches, can be done in a follow-up PR.

Also, OMG our GitHub labels have emojis and consistent coloring :exploding_head:

Carreau:
May I suggest we merge this, see how it goes, and in a couple of weeks write a blog post about our experience ?

rgommers:
@Carreau please do! I'm about to leave on holiday so I'll let you do the honors here (so you can undo them if something isn't quite right).

Carreau:
> please do! I'm about to leave on holiday so I'll let you do the honors here (so you can undo them if something isn't quite right).

Great ! Enjoy time off.

Carreau:
Ok, let's try.

PR #226, opened 2021-07-27
Minor wording fix, use "pyproject.toml-based builds", no PEP xxx

rgommers:
Supporting @pradyunsg in the struggle against using PEP numbers instead of human-understandable terms.


pradyunsg (commented):
❤️

PR #231, opened 2021-08-17
:pencil2: Fix publishing date - move backward

trallard:
This PR fixes the publication date on the blogpost so will be published on 18th Aug 2021


rgommers:
Thanks Tania!

It's also fine to have a time/date slightly in the past, usually that's easier because then the PR merge action can immediately be followed by an announcement/tweet/whatever.

@jaimergp this should show up in ~17 hours from now it looks like.

trallard:
ahhhhhh - good to know I was just bothering Matthias to figure out dates / times  thanks 😉

PR #232, opened 2021-08-19
galleries/team: add Filipe Laíns (FFY00)

FFY00:
:)

Signed-off-by: Filipe Laíns <lains@riseup.net>


galleries/team/metadata.yml
```diff
@@ -164,3 +164,8 @@ title: Saul Shanabrook
 github: saulshanabrook
 projects: ['Vega', 'Jupyter', 'metadsl']
 ---
+name: filipe_lains.png
+title: Filipe Laíns
+github: FFY00
+projects: ['Arch Linux', 'pypa/build', 'CPython Triage', 'importlib_resources']
```
rgommers, line 170: We try to keep this focused on projects that are in scope for Quansight Labs. Of course that boundary is often slightly fuzzy, however I'd say Arch Linux is not in scope. And `CPython Triage` should just be `CPython` I think - you don't have to have commit rights to list a project, regular contributions are enough.

rgommers (commented):
Thanks @FFY00. I suggested a couple of changes to listed projects.

galleries/team/metadata.yml
FFY00: That makes sense. I put it because it demonstrates my experience with packaging, but was not sure.

rgommers: Good point about demo'ing experience. When we talk to people in person we can still mention that of course. For this page the main focus though is things we do as part of Labs.

rgommers (approved):
Thanks @FFY00, LGTM now.

PR #236, opened 2021-08-31
Fix author listing on EOSS cycle 4 blog post

rgommers:
Author list needs to be on a single line, the rest gets cut off.


Carreau:
Ah sorry, my vim auto formatter got overzealous.

PR #237, opened 2021-08-31
Unify author spelling.

Carreau:
Making sure we are using & as separator, using full spelling of author
names with no other characters to make sure this page list authors only
once:

https://labs.quansight.org/authors/

@melissawm Just want to make sure it's the right fix, or do you prefer me to unify as "Melissa Mendonça" ?


melissawm:
Yes, that is perfectly fine. Thanks for asking!

PR #245, opened 2021-09-07
Adds newline to place text under image in Numpy Accessible blog post

thomasjpfan:
Small adjustment to place text under image


trallard:
thank you @thomasjpfan

PR #246, opened 2021-09-08
Automatic handling of pyinotify on mac.

Carreau:
We can tag it as not to be installed.


trallard (commented):
I like this better 👍🏽  merging

PR #247, opened 2021-09-08
Update dependencies.

Carreau:
This bumps some dependencies to latest version.
Ideally we should get both a requirements.txt and requirment.txt.lock
where one has only the actual required libraries and the other all the
pinned versions, so that we can easily update and freeze again, but
that's likely for another time.

I've rebuild locally with both old dependencies and new one to make sure
that the generated content is identical to some variation of the date
and of the XML attribute ordering.


trallard:
looks good to me - merging now

PR #250, opened 2021-09-08
:construction_worker: Add pa11y tests

trallard:
This PR makes a first attempt at adding `pa11y` for accessibility testing to the site

- [x] Adds `package.json` -> npm scripts and devDependencies
- [x] Adds `pa11y` configuration
- [x] Create GitHub action to run `pa11y` tests on push to `main` and `PRs`
- [x] Post results of tests in PR - currently as comments

## Notes:

- I cannot figure why the output formatting is so bad in the PR comments (if anyone has an idea lmk)
- Right now the standard for checks is ` "standard": "WCAG2AA",` which results in very little errors,
if this is changed to ` "standard": "WCAG2AAA",` the results are something like:
```
Running Pa11y on 5 URLs:
 > http://localhost:8000/blog - 77 errors
 > http://localhost:8000 - 5 errors
 > http://localhost:8000/team - 43 errors
 > http://localhost:8000/projects - 8 errors
 > http://localhost:8000/about - 5 errors
```

I do not know what we want to do re standards considering we might have a huge rewrite of the site soon do we want to have WCAG2AAA checks and maybe fixes or does it not make sense  to even look into this right now 🤷🏻‍♀️


Carreau:
That Looks good to me so far ! I"m not too familiar with pa11y, and we can probably configure/work on the issues after the basic setup is done.

rgommers (commented):
Looks like a good idea to add pa11y CI, thanks @trallard!

CI is green but tests failed, so needs a few tweaks still (which is why it's still Draft I guess).

Can we get rid of the bot? CI green/red should be enough, and bots posting comments is never a good idea (it's just extra notification spam).

isabela-pf (commented):
With the limited knowledge I have, this also looks good to me. I'm sure it will be trial and error to figure out details since this is new to most of us.

Building on @rgommers' comment, I'm not seeing the errors appear in the checks section right now. I'm only finding the specifics in the comments. I don't know if there's a way to configure that, but maybe it has to do with [pa11y reporters](https://github.com/pa11y/pa11y-ci#reporter-options)?

trallard:
> CI is green but tests failed, so needs a few tweaks still (which is why it's still Draft I guess).

Yes there are some tweaks needed - but tests will be failing until we tackle all the a11y issues there

Will remove the bot - plus working with the output format of the tests is a super pain

Re  outputs - I tried adding the details in the bot but @rgommers  is right re notification noise, the other way to find the outputs would be to hed directly to the GitHub actions tab and then into the workflow (follow this link for an example: https://github.com/Quansight-Labs/quansight-labs-site/runs/3547260658?check_suite_focus=true#step:10:1)

trallard:
also @isabela-pf  do you have any preference re standard? (as per the initial description?)

trallard:
This should be ready to merge now - I can then go and open a separate PR to tackle the current errors

trallard:
Will merge and have this ready for the sprint Isabela is running today

PR #251, opened 2021-09-09
Quantile Regression post

cgarciae:
Hey! I made this article about Quantile Regression a month ago and posted it on my personal HackMD, got about 3K views. I think this would be very good material for the Quansight Blog.

I use some latex formulas, do we support rendering MathJax?

## Text styling

## Non-text contents

- [ ] If it were to be read as plain text, the blog still makes sense and no information is missing. (they would have to know latex if its not rendered).

### DIFF
--- a/posts/2021/09/quantile-regression.md
+++ b/posts/2021/09/quantile-regression.md
@@ -1,0 +1,425 @@
+# Quantile Regression
+_A simple method to estimate uncertainty in Machine Learning_
+
+Cristian Garcia
+
+<a href="https://github.com/cgarciae/quantile-regression" target="_parent">
+    <img src="https://img.shields.io/badge/R-Open%20Repo-blue"/>
+</a>
+<a href="https://colab.research.google.com/github/cgarciae/quantile-regression/blob/master/main.ipynb" target="_parent">
+    <img src="https://colab.research.google.com/assets/colab-badge.svg" alt="Open in Colab"/>
+</a>
+
+
+## Motivation
+When generating predictions about an output, it is sometimes useful to get a confidence score or, similarly, a range of values around this expected value in which the actual value might be found. Practical examples include estimating an upper and lower bound when predicting an ETA or stock price since you not only care about the average outcome but are also very interested in the best-case and worst-case scenarios in when trying to minimize risk e.g. avoid getting late or not loosing money.
+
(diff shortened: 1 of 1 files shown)


trallard:
HEy @cgarciae  we do support MathJax so you should be good. Some questions:

- Is this ready for review?
- have you checked with @juanis2112  re publication date and adding this to the calendar

TODO:
- this needs to be x-posted to dev.to @cgarciae  ping me for the instructions to join the org

cgarciae:
@trallard this is ready for review. @juanis2112 told me offline she is considering to publishing it in Sep 21.

rgommers (commented):
Hey Cristian, I pushed a commit to make the post render locally, some Nikola metadata was missing. I just scanned the post, so here are some high-level comments:

- It will be unclear to the reader why this blog post is here (on the Labs blog). There is no connection to Labs, or to Elegy beyond the last bullet point of the post.
- What is the main purpose of this post, and who is it aimed at? Is it a tutorial-like thing? Or a new algorithm? Or a way to tell people about Elegy? Or ....?
- The title should reflect that purpose, it's too short now.

tonyfast:
@cgarciae you think you could address some of @rgommers comments to get this is publishable state?

cgarciae:
Thanks @tonyfast for the ping!

@rgommers Its mostly a tutorial for data scientist about a useful but not that popular technique. Elegy and JAX are just used as tools, not the main focus in this case. You make a good point about the content regarding Labs, should I send it to Quansight's Blog instead?

tonyfast:
> You make a good point about the content regarding Labs, should I send it to Quansight's blog instead?

@katebrack @ericdatakelly

cgarciae:
@trallard Should I post it on dev.to regardless?

Katebrack:
@tonyfast Thanks for sending, we can take it for Quansight. Troy and I took a quick pass and think it'd be good. I'll do a deeper read soon and talk to @ericdatakelly about adding it to the pipeline.

ericdatakelly:
@cgarciae you implemented this in our recent client project, correct? (Let's keep the client's name private for now)  For the Quansight blog, we should be able to frame this post in that context.  It would probably only need small changes to the introduction.

cgarciae:
@ericdatakelly yeah, I did some research on this topic to implement it for that client.
As you say, I think the introduction could be improved if we want to add some extra motivations.

cgarciae:
@trallard working with @ericdatakelly and @Katebrack to publish it on the other blog.

PR #275, opened 2021-11-03
Do not use space in directive.

Carreau:
While this is supported by docutils for historic reason it is not in the
spec and make it harder to parse for non-docutils parsers.

It also show the right ways to do things to users.


rgommers (approved):
thanks @Carreau

PR #278, opened 2021-11-10
SciPy and scikits backend extensibility

IvanYashchuk:
Hey all! I'd like to submit this post written together with @rgommers. In this post, we talk about how we'd like to move forward with array interoperability for SciPy, scikit-learn, scikit-image, and other array-consuming libraries.

## Text styling

## Non-text contents


posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,191 @@
...
+* _plum-dispatch_ - implementation of multiple dispatch that follows the ideas from Julia,
+* _uarray_ is a generic backend/multiple dispatch library developed to provide NumPy-independent but similar to NumPy's dispatch functionality.
+
+First two options are very similar in functionality and usability. But the last option, in addition to multiple dispatch, offers granular control using context managers and provides a way to switch backends for the same array type. It's also possible to make the `uarray` dispatcher work without a context manager making the mechanism implicit and similar to other dispatch libraries. Whether we should encourage the implicit registration of backends or not is an open design issue and there's previous discussion on the topic in [SciPy #14266](https://github.com/scipy/scipy/issues/14266). Hameer Abbasi, one of the `uarray` authors, wrote [a blog post](https://labs.quansight.org/blog/2019/07/uarray-update-api-changes-overhead-and-comparison-to-__array_function__/) about the motivation for the `uarray` and how it compares to NumPy's `__array_function__` dispatch protocol.
```
rgommers, line 170: I'll try and tweak this a little I think, since it's been a topic of a lot of messy discussions. Let me try and state why `uarray` is preferred., e.g.:

1. we will have modules where multiple other libraries will want to provide alternatives for the same array type (true for `fft` and `linalg` clearly, perhaps more)
2. we don't want two different systems for the same job (and given 1 above, uarray is the one good candidate)
3. (add simplest example of another feature specific to uarray)

@thomasjpfan may also have some thoughts here, since he is fairly new to uarray and immediately liked it when he wrapped his head around it.

rgommers (commented):
Thanks @IvanYashchuk! Some initial comments, I can do more tomorrow morning.

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,191 @@
...
+
+The ``__array_namespace__`` protocol is:
+* opt-in for users,  
+The new dispatching works only when users choose to use the new Array API compatible modules.
+* explicit,  
+The dispatched functions are determined via the method of the array object user creates.
+* with local control.  
+Which implementation to use is determined by calling methods on the direct arguments of a function.
+
+And the `uarray` dispatcher is more flexible than the ``__array_namespace__`` protocol:
```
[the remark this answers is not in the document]
IvanYashchuk: Agreed, "flexibility" here was referring to the fact that [several alternative design choices](https://numpy.org/neps/nep-0037-array-module.html#appendix-design-choices-for-api-overrides) are possible to implement in the `uarray` case. We can have automatic/implicit registration of a backend, be able to use context managers as well as set global flags.

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+The new dispatching system implemented for SciPy and scikits should:
+* make the API extendable and add the possibility to support alternative array libraries, not just NumPy,
+* provide a way to select backend for the same array type (example: `scipy.fft` + alternative CPU FFT libraries, scikit-learn + [Intel optimized implementation](https://github.com/intel/scikit-learn-intelex)),
+* be able to extend function signature in a backward-compatible way on the base library side (all alternative implementations should not break after the change),
+* have as low overhead as possible.
+
+When designing API override systems, there are many possible design choices. Three major axes of design decisions in the context of NumPy API overrides are outlined in [the appendix of NumPy Enhancement Proposal (NEP) 37](https://numpy.org/neps/nep-0037-array-module.html#appendix-design-choices-for-api-overrides); they are also applicable outside of NumPy:
+* Opt-in vs. opt-out for users,
+* Explicit vs. implicit choice of implementation,
+* Local vs. non-local vs. global control.
```
AnirudhDagar, line 77:
```suggestion
* Opt-in vs. opt-out for users
* Explicit vs. implicit choice of implementation
* Local vs. non-local vs. global control
```

Same

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Goals, wishes, and constraints
+
+Goal: separate the interface from the implementation and let the dispatching system help users and library authors write generic code that works with different kinds of arrays.
+
+The new dispatching system implemented for SciPy and scikits should:
+* make the API extendable and add the possibility to support alternative array libraries, not just NumPy,
+* provide a way to select backend for the same array type (example: `scipy.fft` + alternative CPU FFT libraries, scikit-learn + [Intel optimized implementation](https://github.com/intel/scikit-learn-intelex)),
+* be able to extend function signature in a backward-compatible way on the base library side (all alternative implementations should not break after the change),
+* have as low overhead as possible.
```
AnirudhDagar, line 72:
```suggestion
* make the API extendable and add the possibility to support alternative array libraries, not just NumPy
* provide a way to select backend for the same array type (example: `scipy.fft` + alternative CPU FFT libraries, scikit-learn + [Intel optimized implementation](https://github.com/intel/scikit-learn-intelex))
* be able to extend function signature in a backward-compatible way on the base library side (all alternative implementations should not break after the change)
* have as low overhead as possible
```
Personal inclination to not use "," in these bullets. Not a big deal though. Feel free to ignore if you feel the other way.

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+
+The dispatch using ``__array_namespace__`` method is:
+* opt-in for users,  
+The new dispatching works only when users choose to use the new Array API compatible modules.
+* explicit,  
+The dispatched functions are determined via the method of the array object user creates.
+* with local control.  
+Which implementation to use is determined by calling methods on the direct arguments of a function.
```
AnirudhDagar, line 89:
```suggestion
* opt-in for users:  
The new dispatching works only when users choose to use the new Array API compatible modules.
* explicit:  
The dispatched functions are determined via the method of the array object user creates.
* with local control:  
Which implementation to use is determined by calling methods on the direct arguments of a function.
```

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+* Opt-in vs. opt-out for users,
+* Explicit vs. implicit choice of implementation,
+* Local vs. non-local vs. global control.
+
+## A concrete design proposal
+
+In the figure below, we highlight the parts of the design for implementing the CuPy backend in SciPy, scikit-learn, scikit-image and other core projects which now rely only on NumPy. We are going to use two dispatching mechanisms: one for the modules with compiled code portions, and one for the pure Python functions. The first dispatching mechanism is based on the [`uarray` project](https://uarray.org/), which is a backend dispatcher that allows us to choose the relevant function implementation at runtime. The second dispatching mechanism is based on the ``__array_namespace__`` method from Array API standard, which is a method that allows us to get the Array API implementation module specific to CuPy (or any other array library) at runtime. The ``__array_namespace__`` will be available in NumPy 1.22 and CuPy v10.0.
+
+The dispatch using ``__array_namespace__`` method is:
```
AnirudhDagar, line 83:
```suggestion
The dispatch using ``__array_namespace__`` method is:

```
Same

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+* opt-in for users,  
+The new dispatching works only when users choose to use the new Array API compatible modules.
+* explicit,  
+The dispatched functions are determined via the method of the array object user creates.
+* with local control.  
+Which implementation to use is determined by calling methods on the direct arguments of a function.
+
+And the `uarray` dispatcher is:
```
AnirudhDagar, line 91:
```suggestion
And the `uarray` dispatcher is:

```
Same

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+* opt-out for users,
+* can be implemented to be explicit or implicit,
+* can be implemented to have local, non-local or global control.
```
AnirudhDagar, line 94:
```suggestion
* opt-out for users
* can be implemented to be explicit or implicit
* can be implemented to have local, non-local or global control
```

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+    # Retrieves standard namespace.
+    # Raises if x and y have different namespaces.
+    xp = get_namespace(x, y)
+    out = xp.mean(x, axis=0) + 2*xp.std(y, axis=0)
+    return out
+```
+
+Notice that to make the above code valid, we don't need to import the package that implements mean and std functions explicitly; the required module is stored in x and y variables.
+
+For several concrete use cases of the Array API check out [the dedicated page](https://data-apis.org/array-api/latest/use_cases.html) of the standard. Anirudh Dagar also [wrote a great demo](https://anirudhdagar.ml/array-api-demo/GW_Demo_Array_API.html) recreating gravitational waves tutorial and showcasing what it would take to support PyTorch tensors for a subset of SciPy API using `__array_namespace__` .
```
AnirudhDagar, line 125:
```suggestion
For several concrete use cases of the Array API check out [the dedicated page](https://data-apis.org/array-api/latest/use_cases.html) of the standard. Anirudh Dagar also [wrote a great demo](https://quansight-labs.github.io/array-api-demo/GW_Demo_Array_API.html) recreating gravitational waves tutorial and showcasing what it would take to support PyTorch tensors for a subset of SciPy API using `__array_namespace__` .
```
Fix link to demo.

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+
+## PyData dispatching system
+
+In the future, we'd like to pass alternative arrays to SciPy and scikits functions and get the results without transforming internally to NumPy arrays. Instead, it would automatically handle dispatching to the appropriate implementation, for example, `cupyx.scipy` for GPU-accelerated SciPy functions, scikit-learn-intelex for Intel optimized version of scikit-learn, or cuCIM for GPU-accelerated scikit-image.
+Usually in statically typed languages function overloading is used to implement a function with the same name but arguments of different types. How can we implement something similar to functions overloading but in Python? Well, we could implement one central function checking the types of the arguments using `isinstance()`, or `issubclass()` and calling appropriate implementations based on these checks. This approach doesn't scale well with the number of inputs of different types and is not extendable by third-party implementers without modifying the source code.
```
AnirudhDagar, line 130: I like that we mention this approach here since this is the first thing that comes to mind even though it is not very scalable and efficient. Would it be a good idea to also mention a real world example like the [einops](https://github.com/arogozhnikov/einops) example here since they particularly use this approach to manage backends?

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+cupy_img = cp.array(numpy_img)
+
+print(type(laplace(numpy_img)))
+# <class 'numpy.ndarray'>
+
+print(type(laplace(cupy_img)))
+# <class 'cupy._core.core.ndarray'>
+```
+
+Since Python doesn't provide standard tools for using multiple dispatch there are several dedicated libraries:
```
AnirudhDagar, line 170:
```suggestion
Since Python doesn't provide standard tools for using multiple dispatch there are several dedicated libraries:

```
same

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+* _multimethod_ - pure Python implementation of multiple dispatch with caching of argument types,
+* _plum-dispatch_ - implementation of multiple dispatch that follows the ideas from Julia,
+* _uarray_ is a generic backend/multiple dispatch library developed to provide NumPy-independent but similar to NumPy's dispatch functionality.
```
AnirudhDagar, line 173:
```suggestion
* _multimethod_ - pure Python implementation of multiple dispatch with caching of argument types
* _plum-dispatch_ - implementation of multiple dispatch that follows the ideas from Julia,
* _uarray_ is a generic backend/multiple dispatch library developed to provide NumPy-independent but similar to NumPy's dispatch functionality
```

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+
+## It's already happening
+
+SciPy has GPU and distributed arrays support [on the roadmap](http://scipy.github.io/devdocs/dev/roadmap.html), but adding this support directly to the library would inflate the scope and increase the maintenance burden significantly. scikit-learn has [explicitly stated](https://scikit-learn.org/stable/faq.html#will-you-add-gpu-support) that they cannot afford GPU-specific code in their codebase. A dispatching system is a low-maintenance solution for extending the functionality of the API.
+
+In 2019 Peter Bell added support for backend switching to `scipy.fft` module ([PR #10383](https://github.com/scipy/scipy/pull/10383)) using `uarray`. Now it's possible to tell SciPy to use, for example, CuPy backend, for computing FFT when CuPy's array is passed to functions from `scipy.fft`.
```
AnirudhDagar, line 181:
```suggestion
In 2019, Peter Bell added support for backend switching to `scipy.fft` module ([PR #10383](https://github.com/scipy/scipy/pull/10383)) using `uarray`. Now it's possible to tell SciPy to use, for example, CuPy backend, for computing FFT when CuPy's array is passed to functions from `scipy.fft`.
```

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+
+In this recent blog post, Anirudh Dagar explores array interoperability protocols: `__array_function__` (NEP 18), `__array_namespace__` (NEP 47) and `uarray`. The blog post includes a [section comparing the Array API and `uarray`](https://labs.quansight.org/blog/2021/10/array-libraries-interoperability/#protocol_differences) approaches for interoperability.
```
AnirudhDagar, line 183:
```suggestion
In a recent blog post titled [Array Libraries Interoperability](https://labs.quansight.org/blog/2021/10/array-libraries-interoperability/), Anirudh Dagar explores array interoperability protocols: `__array_function__` (NEP 18), `__array_namespace__` (NEP 47) and `uarray`. The blog post includes a [section comparing the Array API and `uarray`](https://labs.quansight.org/blog/2021/10/array-libraries-interoperability/#protocol_differences) approaches for interoperability.
```
Add link and reword with the title of the blog to follow accessibility guidelines.

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,196 @@
...
+Adding dispatch support will take significant effort to get implemented across different projects. Fortunately, AMD recognizes the problem and is willing to help to fix it by funding Quansight developers to do the initial work.
+
+AMD's vision for the developer community is open, portable tools that enable the same source code to run on different hardware platforms. ROCm is an example of this, supporting both OpenMP and HIP on multiple architectures. "AMD and Quansight are aligned in this direction. We believe that by working together and working with the Python community, we can achieve an open, portable solution for all Python developers," says Terry Deem, AMD Product Manager for ROCm.
+
+<p align="left">
+    <img
+     alt="AMD logo."
+     src="/images/sponsor/AMD_E_Blk_RGB.png">
+    <i></i>
+</p>
```
AnirudhDagar, line 196: This is broken in the current state.

**I'm not sure** but I guess this src path can't be used for adding images to a blog. You will probably need to move the image to `/images/2021/11/` and then change the path accordingly to make this render.

AnirudhDagar (changes requested):
Awesome read @IvanYashchuk. Looks very clear and easy to understand overall.

Especially love the two simple dispatcher figures about how to handle compiled code and pure NumPy, it nicely explains why `__array_function__` or `__array_namespace__` are not enough.

I've made a few comments and made some aesthetic suggestions.

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,363 @@
...
+
+<!-- TEASER_END -->
+
+Over the years array computing in Python has evolved to support distributed
+arrays, GPU arrays, and other various kinds of arrays that work with specialized
+hardware, or carry additional metadata, or use different internal memory
+representations. The foundational library for array computing in the PyData
+ecosystem is NumPy. But NumPy alone is a CPU-only library - and a
+single-threaded one at that - and in a world where it's possible to get a GPU or
```
thomasjpfan, line 43: From my understanding, NumPy with OpenBLAS can take advantage of multiple threads?

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,363 @@
...
+  NumPy-independent but similar to NumPy's dispatch functionality.
+
+First two options are very similar in functionality and usability. But the last
+option, in addition to multiple dispatch, offers granular control using context
+managers and provides a way to switch backends for the same array type. It's
+also possible to make the `uarray` dispatcher work without a context manager
+making the mechanism implicit and similar to other dispatch libraries. Whether
+we should encourage the implicit registration of backends or not is an open
+design issue and there's previous discussion on the topic in [SciPy
+#14266](https://github.com/scipy/scipy/issues/14266). Hameer Abbasi, one of the
```
thomasjpfan, line 316: This is being parsed as a HTML header.

posts/2021/11/pydata-extensibility-vision.md
thomasjpfan: My preference for `uarray` is very pragmatic:

1. SciPy already ships `fft` that integrates with `uarray` and there is momentum for other submodules.
2. Covers use cases that `__array_function__` or `__array_namespace__` does not.

In terms of comparing it to `multimethod` or `plum-dispatch`, I think point 1 already gives `uarray` an edge. Furthermore, `uarray` defines protocols to communicate between API author and backend providers.

posts/2021/11/pydata-extensibility-vision.md
IvanYashchuk: Automatic line wrapping is to blame!

posts/2021/11/pydata-extensibility-vision.md
AnirudhDagar: @IvanYashchuk @rgommers what do you think? I guess this comment got buried under other small review comments.

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,363 @@
...
+
+Since Python doesn't provide standard tools for using multiple dispatch there
+are several dedicated libraries:
+
+* _multimethod_ - pure Python implementation of multiple dispatch with caching
+  of argument types
+* _plum-dispatch_ - implementation of multiple dispatch that follows the ideas
+  from Julia
+* _uarray_ is a generic backend/multiple dispatch library developed to provide
+  NumPy-independent but similar to NumPy's dispatch functionality.
```
peterbell10, line 307: This is okay but a bit vague. How about:
```suggestion
* _uarray_ is a generic backend/multiple dispatch library similar to NumPy's dispatch functionality
   except that the implementation doesn't need to be baked into the array type.
```

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,363 @@
...
+## A concrete design proposal
+
+In the figure below, we highlight the parts of the design for implementing the
+CuPy backend in SciPy, scikit-learn, scikit-image and other core projects which
+now rely only on NumPy. We are going to use two dispatching mechanisms: one for
+the modules with compiled code portions, and one for the pure Python functions.
+The first dispatching mechanism is based on the [`uarray`
+project](https://uarray.org/), which is a backend dispatcher that allows us to
+choose the relevant function implementation at runtime. The second dispatching
+mechanism is based on the ``__array_namespace__`` method from Array API
```
grlee77, line 162:
```suggestion
mechanism is based on the ``__array_namespace__`` method from the Array API
```

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,363 @@
...
+    </i>
+</p>
+
+## Array API dispatching system
+
+Array API is a standard specification of functionality that array libraries need
+to implement. There is a [dedicated test
+suite](https://github.com/data-apis/array-api-tests) to verify the compliance of
+specific implementations with the standard, so it's easy to check all libraries
+adopting the standard to have a unified behavior. NEP 47 has a couple of
```
grlee77, line 202:
```suggestion
adopting the standard have a unified behavior. NEP 47 has a couple of
```

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,363 @@
...
+Notice that to make the above code valid, we don't need to import the package
+that implements mean and std functions explicitly; the required module is stored
+in x and y variables.
+
+For several concrete use cases of the Array API check out [the dedicated
+page](https://data-apis.org/array-api/latest/use_cases.html) of the standard.
+Anirudh Dagar also [wrote a great
+demo](https://quansight-labs.github.io/array-api-demo/GW_Demo_Array_API.html)
+recreating gravitational waves tutorial and showcasing what it would take to
+support PyTorch tensors for a subset of SciPy API using `__array_namespace__`.
```
grlee77, line 233:
```suggestion
support PyTorch tensors for a subset of the SciPy API using `__array_namespace__`.
```

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,363 @@
...
+## PyData dispatching system
+
+In the future, we'd like to pass alternative arrays to SciPy and scikits
+functions and get the results without transforming internally to NumPy arrays.
+Instead, it would automatically handle dispatching to the appropriate
+implementation, for example, `cupyx.scipy` for GPU-accelerated SciPy functions,
+scikit-learn-intelex for Intel optimized version of scikit-learn, or cuCIM for
+GPU-accelerated scikit-image. Usually in statically typed languages function
+overloading is used to implement a function with the same name but arguments of
+different types. How can we implement something similar to functions overloading
```
grlee77, line 244:
```suggestion
different types. How can we implement something similar to function overloading
```

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,363 @@
...
+are several dedicated libraries:
+
+* _multimethod_ - pure Python implementation of multiple dispatch with caching
+  of argument types
+* _plum-dispatch_ - implementation of multiple dispatch that follows the ideas
+  from Julia
+* _uarray_ is a generic backend/multiple dispatch library developed to provide
+  NumPy-independent but similar to NumPy's dispatch functionality.
+
+First two options are very similar in functionality and usability. But the last
```
grlee77, line 309:
```suggestion
The first two options are very similar in functionality and usability. But the last
```

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,363 @@
...
+option, in addition to multiple dispatch, offers granular control using context
+managers and provides a way to switch backends for the same array type. It's
+also possible to make the `uarray` dispatcher work without a context manager
+making the mechanism implicit and similar to other dispatch libraries. Whether
+we should encourage the implicit registration of backends or not is an open
+design issue and there's previous discussion on the topic in [SciPy
+#14266](https://github.com/scipy/scipy/issues/14266). Hameer Abbasi, one of the
+`uarray` authors, wrote [a blog
+post](https://labs.quansight.org/blog/2019/07/uarray-update-api-changes-overhead-and-comparison-to-__array_function__/)
+about the motivation for the `uarray` and how it compares to NumPy's
```
grlee77, line 319:
```suggestion
about the motivation for `uarray` and how it compares to NumPy's
```

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,363 @@
...
+
+SciPy has GPU and distributed arrays support [on the
+roadmap](http://scipy.github.io/devdocs/dev/roadmap.html), but adding this
+support directly to the library would inflate the scope and increase the
+maintenance burden significantly. scikit-learn has [explicitly
+stated](https://scikit-learn.org/stable/faq.html#will-you-add-gpu-support) that
+they cannot afford GPU-specific code in their codebase. A dispatching system is
+a low-maintenance solution for extending the functionality of the API.
+
+In 2019, Peter Bell added support for backend switching to `scipy.fft` module
```
grlee77, line 332:
```suggestion
In 2019, Peter Bell added support for backend switching to the `scipy.fft` module
```

posts/2021/11/pydata-extensibility-vision.md
grlee77: maybe it is just a typo? The folder is named `sponsors` rather than `sponsor`.

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,363 @@
...
+same source code to run on different hardware platforms. ROCm is an example of
+this, supporting both OpenMP and HIP on multiple architectures. "AMD and
+Quansight are aligned in this direction. We believe that by working together and
+working with the Python community, we can achieve an open, portable solution for
+all Python developers," says Terry Deem, AMD Product Manager for ROCm.
+
+<p align="left">
+    <img
+     alt="AMD logo."
+     src="/images/sponsor/AMD_E_Blk_RGB.png">
```
grlee77, line 361:
```suggestion
     src="/images/sponsors/AMD_E_Blk_RGB.png">
```

grlee77 (commented):
This seems like a good overview to me and the provided diagrams are nice. I left a lot of tiny editing nitpicks, but no substantial revisions.

posts/2021/11/pydata-extensibility-vision.md
thomasjpfan: I think I would remove "singled-threaded one at that", so we do not need to go through the details of "when numpy uses BLAS".

posts/2021/11/pydata-extensibility-vision.md
rgommers: Maybe useful indeed, but very briefly. For example "(see `einops.backends` as an example)" (with that being a link).

posts/2021/11/pydata-extensibility-vision.md
rgommers: Only the `linalg` functions that call into OpenBLAS or MKL. Which is only a few handfuls out of O(1000) functions/methods. I think it's fine to say this, NumPy really is single-threaded. It's good at releasing the GIL, but that's about it.

posts/2021/11/pydata-extensibility-vision.md
rgommers: Makes sense to me, adding.

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,363 @@
...
+those array libraries. We want to have a unified design using the same tools
+applicable across SciPy, scikit-learn, scikit-image and other libraries. In this
+situation, it's best to have a discussion spanning multiple projects.
+<!-- This post is the invitation to discuss the idea on this forum https://discuss.scientific-python.org/ -->
+
+
+<p align="center">
+    <img
+     alt="A diagram showing possible interconnections that would be enabled by Array API."
+     src="/images/2021/11/nep-0047-library-dependencies.png">
```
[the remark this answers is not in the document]
rgommers: good catch, thanks. fixed now.

posts/2021/11/pydata-extensibility-vision.md
rgommers: yes indeed, fixed now

posts/2021/11/pydata-extensibility-vision.md
rgommers: Added linking to https://github.com/arogozhnikov/einops/blob/1225f50eebd297f82a3fb97fca87319b522cd99c/einops/_backends.py#L22-L29. See if you agree @IvanYashchuk

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,368 @@
...
+mechanism is based on the ``__array_namespace__`` method from the Array API
+standard, which is a method that allows us to get the Array API implementation
+module specific to CuPy (or any other array library) at runtime. The
+``__array_namespace__`` will be available in NumPy 1.22 and CuPy v10.0.
+
+The dispatch using ``__array_namespace__`` method is:
+
+* opt-in for users
+  The new dispatching works only when users choose to use the new Array API
+  compatible modules.
```
rgommers, line 173: This doesn't seem like a good idea. The opt-in/out approach should work the same for pure Python and compiled code, because that is an implementation detail. This will need discussion on Discourse.

rgommers: I suggest we write it as "One of the key questions to address in this design is whether the behavior should be opt-int or opt-out for users. And if it's enabled by default by a library (and hence opt-out for users, or not even giving users a choice), is it the library which implements the n-D array object that controls this enabling or is it individual array-consuming libraries?"

This was one of the key problems with `__array_function__`. It is a hard question, so I'd like to write it as an actual question like above, and then propose our answer to to it. And reference the uarray auto-registration issue.

So the question then is: what is our proposed answer? Neither of the two extremes seem appealing:
- all opt-out we tried with `__array_function__`, was a blocker for e.g. JAX and PyTorch
- a very granular opt-in per (e.g.) SciPy submodule will be too cumbersome

So the best options seem to be:
- opt-in per package, or
- a global opt-in per array library

posts/2021/11/pydata-extensibility-vision.md
AnirudhDagar: Yes, I earlier tried to fix the path but that's not the issue. You can still try to see, the logo is never rendered. You need to move it into `/images/2021/11/` to use in the blog.

rgommers: No that's not the case, it works as is. Many other blog posts have images in `images/...` rather than in `posts/`.

AnirudhDagar: Interesting. Below is a screenshot of where the logo is supposed to be, on the rendered website locally.
image

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,388 @@
...
+making the mechanism implicit and similar to other dispatch libraries. Whether
+we should encourage the implicit registration of backends or not is an open
+design issue and there's previous discussion on the topic in
+[SciPy #14266](https://github.com/scipy/scipy/issues/14266). Hameer Abbasi, one of the
+`uarray` authors, wrote
+[a blog post](https://labs.quansight.org/blog/2019/07/uarray-update-api-changes-overhead-and-comparison-to-__array_function__/)
+about the motivation for `uarray` and how it compares to NumPy's
+`__array_function__` dispatch mechanism.
+
+## It's already happening
```
rgommers, line 344: I'm inclined to change this title to `Next steps`. Most of the content below is already reflected higher up. What we should end with I think is a rough outline of the next steps:
- discuss on Discourse to align with maintainers of SciPy, scikit-learn, scikit-image and the wider community;
- come to a conclusion on a few key points, in particular the opt-in vs. opt-out topic
- merge some of the uarray backends, and then explore using the array API standard in SciPy and scikit-learn with CuPy 10.0

And then a concluding statemen, like "This is an exciting project, we look forward to ... <etc."

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,388 @@
...
+GPU, as well as running in distributed mode across multiple GPUs. In this blog
+post, we talk about the motivation behind the proposal and describe our vision
+for making array libraries more interoperable.
+
+With the proposed dispatch layers, the downstream libraries would be enabled to
+use multiple kinds of arrays, without having to have a hard dependency on all of
+those array libraries. We want to have a unified design using the same tools
+applicable across SciPy, scikit-learn, scikit-image and other libraries. In this
+situation, it's best to have a discussion spanning multiple projects.
+<!-- This post is the invitation to discuss the idea on this forum https://discuss.scientific-python.org/ -->
```
rgommers, line 26: I still need to look at this first two paragraphs.

posts/2021/11/pydata-extensibility-vision.md
rgommers: Odd. Works for me:

image

Maybe a difference in versions, not sure what's going on here.

This is the same as in many other blog posts, it really should work. Is it broken for you on other posts? There are a ton of images in `images/` and `images/sponsors/`.

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,388 @@
...
+
+We need two dispatching mechanisms: one for the modules with compiled code
+portions, and one for pure Python code.
+The first dispatching mechanism is based on the [`uarray`
+project](https://uarray.org/), which is a backend dispatcher that allows us to
+choose the relevant function implementation at runtime. The second dispatching
+mechanism is based on the ``__array_namespace__`` method from the Array API
+standard, which is a method that allows us to get the Array API implementation
+module specific to CuPy (or any other array library) at runtime.
+``__array_namespace__`` will be available in NumPy 1.22 and CuPy v10.0.
```
rgommers, line 173: The second design question to point out is the gray area of dispatching: when something is mostly pure Python code that we want to reuse, _but_ there's some compiled code somewhere. That may be private API, but it needs dispatching still.

rgommers: I thought about it, and we do need to mention it - because it's a fairly obvious gap to maintainers reading this post. I'll have a go at formulating something to address this.

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,388 @@
...
+## Goals, wishes, and constraints
+
+Goal: separate the interface from the implementation and let the dispatching
+system help users and library authors write generic code that works with
+different kinds of arrays.
+
+The new dispatching system implemented for SciPy and scikits should:
+
+* make the API extendable and add the possibility to support alternative array
+  libraries, not just NumPy
```
rgommers, line 136: Another thing just came to mind. What should also be supported is "array duck types", i.e. things that are not subclasses of the base array object in some library, but are compatible with it and use that library's API and implementations. For example, objects that implement `__array_function__` but don't subclass `numpy.ndarray`, or implement `__torch_function__` but don't inherit from `torch.Tensor`.

That will work fine with `uarray` (because it just tries all registered backends and let's them say if they accept the object or not), but it doesn't work with any of the multiple dispatch implementations AFAIK.

posts/2021/11/pydata-extensibility-vision.md
IvanYashchuk: Thank you, Anirudh, for the suggestion, and thank you, Ralf, for adding the link.

posts/2021/11/pydata-extensibility-vision.md
IvanYashchuk: Yes, that's what I meant, OpenBLAS/MKL is not a part of NumPy, it's a dependency that can be multithreaded.

posts/2021/11/pydata-extensibility-vision.md
IvanYashchuk: I think with any multiple dispatch implementation new unknown objects should be directed to the default implementation and then accepted or rejected. In types this means that the default implementation accepts `typing.Any` or `object`.
That works fine with `uarray` because it kind of overwrites the default implementation.

Yes, we should add this to the list of requirements.

posts/2021/11/pydata-extensibility-vision.md
IvanYashchuk: Right, listing the next steps is a better way to end this blog post.

posts/2021/11/pydata-extensibility-vision.md
IvanYashchuk: I agree that it doesn't feel right to have different enabling/registration mechanisms for pure Python or compiled code. It would be confusing and difficult to remember what to use when. I agree that writing it as a question is better.

In the case of Array API individual array-consuming libraries need to choose how they handle inputs both with and without the  `__array_namespace__` method. Using `xp = input.__array_namespace__()` instead of `np = numpy` is not the only possible choice to implement support for Array API compliant libraries. Checking for `__array_namespace__` could be put inside an if statement using global flag, or a context manager.

Maybe the backend dispatching using `uarray` should be enabled only for array objects with `__array_namespace__` method. For other arrays then fall back to the generic implementation coercing to NumPy. Should the context manager (`with uarray.set_backend(...)`) be allowed in this case possibly ignoring absence of `__array_namespace__`?

I think we should encourage auto-registration of `uarray` backends upon import, but the dispatching itself should not be turned on by default on the array-consuming library side. The same switch should control enabling `uarray` dispatching and use of `__array_namespace__` simultaneously to have consistent behavior.

posts/2021/11/pydata-extensibility-vision.md
rgommers: > I think with any multiple dispatch implementation new unknown objects should be directed to the default implementation and then accepted or rejected.

I disagree I think. That effectively means you're going to do the wrong thing for duck arrays (example, [pytorch/nestedtensor](https://github.com/pytorch/nestedtensor/blob/b050671e2c3618d0ff15f468922d9ea514b0c695/nestedtensor/nested/nested.py#L200)). It should either try the default implementation last (which is what `uarray` does), or it should be a principled type based dispatch, which means all unknown objects should raise immediately without dispatch happening. Otherwise it becomes a mess where numpy-likes are accepted and other-array-likes are accepted too but end up doing being converted to numpy arrays rather than being treated by the correct library.

posts/2021/11/pydata-extensibility-vision.md
rgommers: > Checking for __array_namespace__ could be put inside an if statement using global flag, or a context manager.

Yes, that's a good point.

> I think we should encourage auto-registration of `uarray` backends upon import, but the dispatching itself should not be turned on by default on the array-consuming library side. The same switch should control enabling `uarray` dispatching and use of `__array_namespace__` simultaneously to have consistent behavior.

I think I like this idea.

posts/2021/11/pydata-extensibility-vision.md
rgommers: Let's leave this as is, should be okay.

posts/2021/11/pydata-extensibility-vision.md
rgommers: @IvanYashchuk will have a go at this one.

posts/2021/11/pydata-extensibility-vision.md
rgommers: Added:
```
+* support "array duck types", i.e. custom array objects that are compatible
+  with a particular array library via (for example) `__array_function__` or
+  `__torch_function__`.
```

posts/2021/11/pydata-extensibility-vision.md
IvanYashchuk: Added in https://github.com/Quansight-Labs/quansight-labs-site/pull/278/commits/2e77cdc415c9df34d52d4be478b9e3335401f91d.
Once we have direct links to the discussions we should update the text here.

posts/2021/11/pydata-extensibility-vision.md
IvanYashchuk: I removed the current characterization of `uarray` and `__array_namespace__` as it's up to an upcoming discussion to decide what design options to pursue.
I added the proposed question to the text.

rgommers: I like how this reads now. Thanks!

posts/2021/11/pydata-extensibility-vision.md
rgommers: Looks good too. Pushed some tiny tweaks, overall it read quite well.

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,403 @@
...
+GPUs.
+
+We feel like a shared vision has emerged, in bits and pieces. In this post, we
+aim to articulate that vision and suggest a path to making it concrete,
+focusing on three libraries at the core of the PyData ecosystem: SciPy,
+scikit-learn and scikit-image. We are also happy to share that AMD has
+recognized the value of this vision, and is partnering with Quansight Labs to
+help make it a reality.
+
+<!-- TEASER_END -->
```
rgommers, line 34: Updated these first three paragraphs, please have a look @IvanYashchuk

rgommers: Also merged in `main`, because the styling changed and I wanted to know where to put `TEASER_END`

posts/2021/11/pydata-extensibility-vision.md
rgommers: Addressing this in a paragraph below the hand-drawn dispatcher diagrams.

posts/2021/11/pydata-extensibility-vision.md
IvanYashchuk: I am happy with the changes! 
Preview now works which is great!

posts/2021/11/pydata-extensibility-vision.md
```diff
@@ -0,0 +1,410 @@
...
+.. title: A vision for extensibility to GPU & distributed support for SciPy, scikit-learn, scikit-image and beyond
+.. slug: pydata-extensibility-vision
+.. date: 2021-11-15 10:00:00 UTC-07:00
+.. author: Ivan Yashchuk, Ralf Gommers
+.. tags: SciPy, NumPy, CuPy, scikit-learn, scikit-image, uarray, Array API
+.. category:
+.. link:
+.. description:
+.. type: text
+.. previewimage: /images/2021/11/nep-0047-library-dependencies.png
```
IvanYashchuk, line 11: I'm not sure how whether it's supposed to work, but this field does nothing at the moment.

rgommers: I believe that you'll only see that in things like the Twitter link preview thumbnail.

posts/2021/11/pydata-extensibility-vision.md
rgommers: Great, thanks:)

rgommers:
In it goes 🤞🏼. Thanks @IvanYashchuk, @peterbell10, @grlee77, @thomasjpfan, @AnirudhDagar!

rgommers:
We're writing a post to start the discussion on the SPEC Discourse now, then updating the link in this post, then a tweet, then post internally in Slack.

PR #289, opened 2022-01-07
IPython 8

Carreau:
## Text styling

## Non-text contents

### DIFF
--- a/posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
+++ b/posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
@@ -8,12 +8,12 @@
 .. link:
 .. description:
 .. type: text
-.. previewimage: 
+.. previewimage:
 -->
 
 This is a companion post from the [Official release of IPython 8.0], that describe what we leaned with this large new
 major IPython release. We hope it will help you apply best practices, and have an easier time maintaining your projects,
-or helping other. Trust me your future self will thank you. 
+or helping other. Trust me your future self will thank you.
 
 We'll focus on many patterns that made it easier for us to make IPython 8.0 what it is with minimal time involved.
 
@@ -28,16 +28,16 @@ Make the codebase faster to navigate and simpler to understand.
 As Brian Kernighan put it: “Debugging is twice as hard as writing the code in the first place. Therefore, if you write
 the code as cleverly as possible, you are, by definition, not smart enough to debug it.”
 
-So I'd like to keep my code as simple as possible with less states or condition to keep in mind. 
+So I'd like to keep my code as simple as possible with less states or condition to keep in mind.
 
 Seen it in another way, your software progressing both because it's leading edge is moving forward, as it does because
-the trailing end is also catching up. And for this release of IPython we focused on both. 
+the trailing end is also catching up. And for this release of IPython we focused on both.
 
 
 ## LBYL vs EAFP
 
 Look Before You Leap and Easier to Ask Forgiveness than Permissions are two practices each with their advantages and
-inconvenient. Python tend the prefer the second one as try/except is "easy" in Python. 
+inconvenient. Python tend the prefer the second one as try/except is "easy" in Python.
 
 This first implementation will often feel more idiomatic Python:
 
@@ -76,11 +76,11 @@ Explicit is better than implicit.
 ...
 ```
 
-So my first tip: Always avoid catching  `ImportError`s when you can compare version numbers. 
+So my first tip: Always avoid catching  `ImportError`s when you can compare version numbers.
 
- - having two import lines will not make import slower. 
+ - having two import lines will not make import slower.
  - it's explicit.
- - it's easier to search for. 
+ - it's easier to search for.
  - it's easier to remove.
 
 
@@ -100,7 +100,7 @@ except ImportError:
 
 Which adds a fallback for [numpy version older than numpy 1.3 from 2008](https://github.com/numpy/numpy/commit/ba9a02dcb2c3ca635076a75cc9eb0f406e00ceed).
 It tooks me ~30 minutes to find this informations, which could have been seconds would the author (which could have been
-me) had checked version.  A proper version check would also had this code removed years ago. 
+me) had checked version.  A proper version check would also had this code removed years ago.
 
 While IPython has few dependencies beyond Python, we do so with Python version itself, and always compare with
 `sys.version_info` made it straightforward to find all dead code once we bumped or minimal version to 3.8+.
@@ -113,12 +113,12 @@ native to Python. Many simplification leading do even more down the line up to c
 
 Warnings in general are much more powerful and complex than at first look. Once you understand them well they are an
 extremely powerful features. When use correctly they give you an expressive way to communicate with your
-users and dependees as well as you in the future. Misused they can be just noise that trains everybody to ignore them. 
+users and dependees as well as you in the future. Misused they can be just noise that trains everybody to ignore them.
 
-Here is a quick tip/summary of what this section will expand upon. 
+Here is a quick tip/summary of what this section will expand upon.
 
- - Always set `warnings.warn(stacklevel=...)` to the right value (at least 2). 
- - Be descriptive of what "deprecated" means. 
+ - Always set `warnings.warn(stacklevel=...)` to the right value (at least 2).
+ - Be descriptive of what "deprecated" means.
  - Be descriptive of what the replacement is.
  - Always indicate since when it is deprected / the replacement is available.
  - Don't be afraid to use multiple line strings.
@@ -127,7 +127,7 @@ Here is a quick tip/summary of what this section will expand upon.
 ## Always use `stacklevel=...`
 
 
-Setting the stacklevel ensure that python reports the right place where the deprecated feature is used. 
+Setting the stacklevel ensure that python reports the right place where the deprecated feature is used.
 
 ### Make it easier to fix
 
@@ -154,11 +154,11 @@ While setting `stacklevel=2` will point to the right file, line, and show the pr
 ```
 
 Most terminal and editor now recognize the `file/path.py:LineNumber` syntax, so on my setup it is also a single click to
-open the right file on the right location and fix it. 
+open the right file on the right location and fix it.
 
 ### makes it easier to find
 
-Most test runners have options to turns DeprecationWarnings into error **only in the code you maintain**. 
+Most test runners have options to turns DeprecationWarnings into error **only in the code you maintain**.
 
 [napari] for example uses the  following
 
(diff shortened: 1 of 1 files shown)


posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
```diff
@@ -0,0 +1,303 @@
...
+
+
+Setting the stacklevel ensure that python reports the right place where the deprecated feature is used. 
+
+### Make it easier to fix
+
+```
+# file example.py
+import warnings
+def fucntion(argument=None):
```
echarles, line 137:
```suggestion
def function(argument=None):
```

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
```diff
@@ -0,0 +1,303 @@
...
+This first implementation will often feel more idiomatic Python:
+
+```python
+from pahtlib import Path
+
+p = Path('test')
+
+try:
+   p.mkdir()
+Except FileExistsError:
```
thomasjpfan, line 51:
```suggestion
except FileExistsError:
```

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
```diff
@@ -0,0 +1,303 @@
...
+.. author: Matthias Bussonnier
+.. tags: IPython, Software, Open-Source, best-practice
+.. category:
+.. link:
+.. description:
+.. type: text
+.. previewimage: 
+-->
+
+This is a companion post from the [Official release of IPython 8.0], that describe what we leaned with this large new
```
thomasjpfan, line 14:
```suggestion
This is a companion post from the [Official release of IPython 8.0], that describe what we learned with this large new
```

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
```diff
@@ -0,0 +1,303 @@
...
+        pass
+```
+
+
+Which adds a fallback for [numpy version older than numpy 1.3 from 2008](https://github.com/numpy/numpy/commit/ba9a02dcb2c3ca635076a75cc9eb0f406e00ceed).
+It tooks me ~30 minutes to find this informations, which could have been seconds would the author (which could have been
+me) had checked version.  A proper version check would also had this code removed years ago. 
+
+While IPython has few dependencies beyond Python, we do so with Python version itself, and always compare with
+`sys.version_info` made it straightforward to find all dead code once we bumped or minimal version to 3.8+.
```
thomasjpfan, line 106: The above example has to deal with `numpy`. I think readers would be interested in how you are comparing numpy versions now. This can be done either with a link to IPython code or a short code snippet.

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
```diff
@@ -0,0 +1,303 @@
...
+~/foo.py:3: UserWarning: `argument is deprecated`
+  example.function(1)
+```
+
+Most terminal and editor now recognize the `file/path.py:LineNumber` syntax, so on my setup it is also a single click to
+open the right file on the right location and fix it. 
+
+### makes it easier to find
+
+Most test runners have options to turns DeprecationWarnings into error **only in the code you maintain**.
```
thomasjpfan, line 161: I think turning `DeprecationWarnings` warnings into errors for your libraries dependencies is also very valuable. (For example, when NumPy is issuing deprecation warnings and your library depends on NumPy)

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
```diff
@@ -0,0 +1,303 @@
...
+## Communication and Explicitness are keys
+
+
+You want to be explicit to your users and future self. All the explicit
+informations will make it easier for you in the long term.
+
+Especially in the open-source and volunteer work where time is scarce, you want
+to carefully manage where it is spent. 
+
+A waring is likely going to be updated only a couple of time, but may be seen
```
thomasjpfan, line 288:
```suggestion
A warning is likely going to be updated only a couple of time, but may be seen
```

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+
+Some authors also like to add in the deprecation the version where the
+functionality will be removed. I tend to avoid as:
+
+ - I believe this gives users explicit authorisation to delay updating API.
+   Though you as a maintainer want to get rid of it as fast as possible.
+ - I often came across deprecation that were reverted/delayed/not removed. This
+   becomes  confusing to users, and can lead to mistrust of deprecation warning.
```
thomasjpfan, line 260: These are interesting points for not including the version the functionality will be removed.

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
```diff
@@ -0,0 +1,303 @@
...
+I thus prefer to stay factual, and I don't claim to predict the future.
+
+
+### Use multiline strings
+
+Python has multiline strings with triple backticks, they are not limited to
+docstrings. Of course you can (and should) mention deprecation in docstrings,
+but you can and should also use multiline strings in warnings messages. 
+
+The more you break the back of the work for your users, the more likely they are
```
thomasjpfan, line 271: I am confused about what "break the back of the work for your users" means in this context. Is there another way to phrase it?

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
```diff
@@ -0,0 +1,303 @@
...
+.. category:
+.. link:
+.. description:
+.. type: text
+.. previewimage: 
+-->
+
+This is a companion post from the [Official release of IPython 8.0], that describe what we leaned with this large new
+major IPython release. We hope it will help you apply best practices, and have an easier time maintaining your projects,
+or helping other. Trust me your future self will thank you.
```
thomasjpfan, line 16: "Trust me your future self will thank you" feels like a promise into the future and runs counter to your statement about deprecations at the end:

> I thus prefer to stay factual, and I don't claim to predict the future.

My preference is to remove it and to combine this paragraph with "We'll focus on many patterns".

thomasjpfan (commented):
Thanks for the post @Carreau !

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
Carreau: Good point, I'm not comparing numpy version anymore as we just say in setup.py

```
 test_extra=[
...
        "numpy>=1.19",
...
    ],
```

But I'll extend.

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
Carreau: Ah but what I am trying to convey here is the same, if numpy `stacklevel` is set properly turrning all DeprecationWarnigns in napari code into errors, will let you know that you need to update your code, while still letting numpy use deprecated functionalities internally.

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
Carreau: Does most of the work, or give as much instructions so that they have as little as possible to do. I wasn't sure I I know how to say it in french and had to google expression to try to convey it.

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
Carreau: +1

Carreau:
Thanks @thomasjpfan , I've updated the sections you commented on.

posts/2022/01/ipython-8.0-lessons-learned-maintaining-software.md
```diff
@@ -0,0 +1,379 @@
+<!--
+.. title: IPython 8.0, Lessons learned maintaining software
+.. slug: ipython-8.0-lessons-learned-maintaining-software
+.. date: 2022-01-10 02:51:58 UTC-05:00
```
Carreau, line 4:
```suggestion
.. date: 2022-01-12 14:00:00 UTC+01:00
```

ccordoba12:
I don't understand: why this one was merged before the final IPython 8 release?

Carreau:
Because it was uploading while I merged. 

https://pypi.org/project/ipython/8.0.0/

ccordoba12:
Ups, sorry about that. Since I saw a placeholder link for a release announcement at the Jupyter blog, I thought this could be a mistake.

Carreau:
I also needed the URL to be live to update https://blog.jupyter.org/release-of-ipython-8-0-6e034ff122ef and link to it, this other blog post was published at the "same" time.

rgommers:
Nice post Matthias!

Carreau:
> Ups, sorry about that. Since I saw a placeholder link for a release announcement at the Jupyter blog, I thought this could be a mistake.

Yeah it's a bit of a chicken and egg problem, I need to publish and update after, and it takes time to deploy. I think I should be able to guess the final URL but I was a bit lazy. It should be fixed in the source now, but will take time to hit CDN.





> Nice post Matthias!

Thanks ! Any chance for numpy to grow a `version_info` tuple ?

rgommers:
> Thanks ! Any chance for numpy to grow a `version_info` tuple ?

Depends, for what? There's a bunch of flavors in `np.version`, but that's a weird thing. The only thing that was meant to be public was `np.__version__`.

There's no standard for a tuple version; the string can adhere to PEP 440. So is a tuple useful?

Carreau:
> So is a tuple useful

It's much easier to compare, especially since tuple don't need to be the same length:

```
>>> (3,) < (7,3,'7rc')
True
```

Using a version string is tricky as I pointed in the post as splitting on dots and converting to int may fail.
For me at least a tuple that has major/minor as numeric values.

The following would be cool for me
```
version_info = packaging.version.parse(numpy.__version__).release
```

Also numpy is taken as an example by many other project in the scientific ecosystem.
Or at least if not a tuple, expose a version object that properly compare itself to tuples, so that dependees don't roll their own parsing logic.

rgommers:
Okay, we can add it I think. As a plain tuple of 3 integers. I'd prefer it be called `version_tuple` rather than `version_info`. Wanna submit a PR?

Carreau:
> Okay, we can add it I think. As a plain tuple of 3 integers. I'd prefer it be called `version_tuple` rather than `version_info`. Wanna submit a PR?

https://github.com/numpy/numpy/pull/20803

PR #302, opened 2022-02-02
Blog post: Making GPUs accessible to PyData ecosystem

aktech:
## Text styling

## Non-text contents


posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,486 @@
...
+plt.show()
+```
+
+Let's walk through the code above to get a sense of what we're trying to achieve here.
+
+- We have a dataset of greek coins from pompeii (imported from `skimage.data.coins`)
+- We resize the image to 20% to speed up processing
+- Convert the image to graph data structure
+- Apply spectral clustering on the graph (via `kmeans` and `discretize` labelling).
+- Plot the resulting clustering
```
IvanYashchuk, line 281: I think it would be helpful to have the sections enumerated and apply the same enumeration to the comments in the code snippet above.
So `5. Plot the resulting clustering` and `# 5. Plotting`.

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,486 @@
...
+image after rescaling it to a certain proportion, for this demo rescaled the image
+to 0.1, 0.2, 0.4, 0.6, 0.8 and 1x.
+
+
+#### On AMD GPU:
+
+![NumPy cs CuPy AMD](/images/2022/02/numpy_vs_cupy_amd.png)
+
+The plot for AMD GPU is not what you would expect, the computation is faster
+with numpy (i.e. on CPU) compared to cupy (i.e. on GPU) for all
```
IvanYashchuk, line 415: Let's make sure that we are using "NumPy" and "CuPy" throughout the post.
```suggestion
with NumPy (i.e. on CPU) compared to CuPy (i.e. on GPU) for all
```

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,486 @@
...
+image sizes, this is due to the slow device synchronization issue on AMD GPUs.
```
IvanYashchuk, line 416: ... which we are investigating and will submit a bug report to the CuPy project to discuss the strategies for fixing the problem.

rgommers: That's not worth tweaking I think - we need to fix or work around the issue before we can publish this blog post, right?

aktech: We have not been able to isolate the synchronization issue based on my little understanding of AMD and cupy. I am happy to spend more time on this, if we're able to?

rgommers: @aktech let's discuss on Slack, I'll summarize there in a minute.

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,481 @@
...
+## Results
+
+Here we will talk about about the performance of spectral clustering on GPU vs CPU
+
+We ran this on AMD as well as NVIDIA GPUs and here is a plot of the performance.
+The x-axis of the plot is the image dimension, i.e the dimension of the original
+image after rescaling it to a certain proportion, for this demo rescaled the image
+to 0.1, 0.2, 0.4, 0.6, 0.8 and 1x.
+
+
```
rgommers, line 411: Do you understand why smallest image sizes are slower on both AMD and NVIDIA GPUs? Is that real, or some kind of warm-up effect of the benchmarks?

aktech: I don;t have a way to prove it, but as per my understanding on GPUs this is due to:
https://github.com/Quansight-Labs/quansight-labs-site/pull/302/files#diff-fa53d9a3136fe7fcd2fe2137246cd22993c1a5a69a9fb9af285e41663e3c3b3cR425-R426

rgommers: That explains why GPU is slower than CPU, but I don't think it can ever explain why a smaller image size takes more than a larger size on the same GPU.

rgommers: It's possible it's a real effect of course (something happens to work out better with cache size or an algorithm is switching behavior above a certain size). 

It's also possible that it's an artifact. For example, if you add an even smaller image size as the first data point, and then discard that new data point from the benchmarks, does the effect go away?

aktech: Ah, yes that's true I misinterpreted your question. I don't have an idea about that.

aktech: >  For example, if you add an even smaller image size as the first data point, and then discard that new data point from the benchmarks, does the effect go away?

That's a good point, let me try that.

rgommers: That looks much nicer and more sensible. There's still some noise, but overall this matches the expected behavior I'd say.

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,481 @@
+<!--
+.. title: Making GPUs accessible to PyData Ecosystem via Array API
```
rgommers, line 2: This title feels just slightly off. If I'd keep the message unchanged, it'd be:
_Making GPUs accessible to the the PyData Ecosystem via the Array API Standard_.

Which is very long. An alternative would be _Making GPUs accessible to the PyData Ecosystem_. Which is maybe a bit too generic. So I'd lean towards the first one. Thoughts?

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,481 @@
...
+
+
+
+GPUs have become an essential part of the scientific computing stack and with
+the advancement in the technology around GPUs and the ease of accessing a GPU
+on cloud or on-prem, it is in the best interest of the PyData community to spend
+time and effort to make GPUs accessible for the users of PyData libraries. A
+typical user in the PyData ecosystem is quite familiar with the APIs of libraries
+like scipy, scikit-learn, scikit-image and at the moment all these
+libraries only support single core operations on CPU. In this blog post will
```
rgommers, line 22: I'll rephrase this to reflect that there are parts (`linalg`, `multiprocessing`, scikit-learn's OpenMP usage) are multi-threaded. We can add a footnote to explain that better perhaps.

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,481 @@
...
+
+The motivation for this exercise comes from the fact that having GPU-specific code
+in SciPy and scikits will be too-specialized and very hard to maintain. The ultimate
+goal is to have GPU support in SciPy, scikit-learn and scikit-image using CuPy. Since
+CuPy has support for NVIDIA GPUs as well as AMD’s ROCm GPU platform, this demo will
+work on either of those GPUs.
+
+### A primer on Array API Standard
+
+![Data APIs logo](/images/2022/02/data-apis-logo.png)
```
rgommers, line 44: I'm inclined to remove this logo. It feels out of place, and very few people will recognize it as that logo anyway.

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,481 @@
...
+adoption mechanism in the [NEP 47](https://numpy.org/neps/nep-0047-array-api-standard.html)
+it is now possible to write code that is portable between NumPy and other array/tensor
+libraries that adopt the Array API Standard. We will also discuss the workflow and
+challenges for achieving the same.
+
+<!-- TEASER_END -->
+
+## Motivation and Goal
+
+The motivation for this exercise comes from the fact that having GPU-specific code
```
rgommers, line 36: In this section (or the intro) we should cross-link to the announcement blog post for the AMD-sponsored project that this demo work is part of.

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,481 @@
...
+in SciPy and scikits will be too-specialized and very hard to maintain. The ultimate
+goal is to have GPU support in SciPy, scikit-learn and scikit-image using CuPy. Since
```
rgommers, line 38: Note that using CuPy is not quite the ultimate goal. Rather, we want to enable these libraries to accept any type of array, so they can work with CuPy, PyTorch, JAX, Dask and other libraries. Then hardware vendors like AMD can focus on support for their own type of hardware.

More importantly, there's a mix here between the goal of the larger project and the goal of this demo. Describing both would be good.

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,481 @@
...
+Source: [https://www.brooklynmuseum.org/opencollection/archives/image/51611](https://www.brooklynmuseum.org/opencollection/archives/image/51611)
+
+## Process
+
+#### `get_namespace` - getting the array API module
+
+The [NEP 47](https://numpy.org/neps/nep-0047-array-api-standard.html) (Adopting
+the array API standard) outlines a basic workflow for adopting array API
+standards for array provider and consumer libraries. Now we will describe
+in detail about how we went through the process of implementing this.
```
rgommers, line 107: @aktech should this sentence say _"of implementing this in forks of SciPy, scikit-learn and scikit-image."_?

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,484 @@
...
+generic enough to be defined in the Array API or we haven't got around implementing those.
+
+Some examples of unimplemented functions that we encountered during this excercise:
+
+
+* `np.atleast_3d`
+* `np.strides`
+* `np.real`
+* `np.clip`
+
```
rgommers, line 361: Here we need a line concluding what was done to resolve the challenge. Was it a matter of rewriting the code in each case @aktech?

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,484 @@
...
+
+# Doesn't Works with Array API
+In [5]: z = npx.arange(10)
+In [6]: z[npx.arange(2)]
+-----------------------------------------------------------------
+...
+IndexError: Non-zero dimensional integer array indices are not allowed in the array API namespace
+```
+
```
rgommers, line 404: This challenge section should probably conclude with something like: _"Each of the inconsistencies can be worked around. Renames may require some detective work but are typically easy code changes. Differences in behavior can be harder to deal with. The Array API standard aims to provide a set of functionality that's orthogonal but complete - so if it's possible to perform some array operation on accelerator devices, then there's typically a way to access it in the standard."_

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
rgommers: https://labs.quansight.org/blog/2021/11/pydata-extensibility-vision/

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,497 @@
...
+- `scikit-learn`
+- `scikit-image`
+
+The changes are present in the `array-api-gpu-demo` branch of [@aktech](https://github.com/aktech)'s fork
+of the above projects.
+
+In the above mentioned repo, you can find the code with which we created Docker
+images for both NVIDIA & AMD platforms. The README also contains instructions
+for how to obtain the built Docker images from Docker Registry. Once you are
+inside the Docker container, running the demo is as simple as running:
```
rgommers, line 458: I rephrased this bit to make it a little more explicit to people (like myself) who are not 100% fluent speaking Docker lingo. Please double check it all reads well @aktech

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,497 @@
...
+## Acknowledgment
+
+This demo work was part of a larger project to make PyData libraries work with
+multiple array libraries and thereby run on GPUs. This is a significant
+undertaking, and this work would not have been possible without funding
+from AMD for a team at Quansight Labs.
+
+Special thanks to [Ivan Yashchuk](https://github.com/IvanYashchuk) and
+[Ralf Gommers](https://github.com/rgommers) for their guidance throughout the
+project.
```
rgommers, line 489: I appreciate the thought @aktech, but I'm thinking it may be preferred to leave this acknowledgement out. Other team members contributed as well to this work, and the main purpose of this section is funding acknowledgement. 

@IvanYashchuk WDYT?

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
aktech: The first one sounds reasonable to me.

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
aktech: For some of them it was reimplementing an equivalent function and for remaining it was a hack basically, which is calling those functions on `._array` attribute and then converting back to array api.

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
IvanYashchuk: Let's keep only funding acknowledgement.

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,507 @@
...
+The motivation for this exercise comes from the fact that having GPU-specific code
+in SciPy and scikits will be too-specialized and very hard to maintain. The goal of
+this demo is to (a) to show things working in a real-world example on GPUs, and
+(b) demonstrate that there is a significant performance benefit. We'll be using
+CuPy as the GPU library for this demo. Since CuPy has support for NVIDIA GPUs
+as well as AMD’s ROCm GPU platform, this demo will work on either of those GPUs.
+
+The ultimate goal is to have SciPy, scikit-learn and scikit-image accept any type
+of array, so they can work with CuPy, PyTorch, JAX, Dask and other libraries.
+This project is part of a larger [vision sponsored by
```
rgommers, line 45: I tried to use "demo" and "work" for the Greek coins demo in this blog post, and "project" for the overall project sponsored by AMD in a consistent way. So I'd like to give this sentence a tweak like that. Other than that it reads well.

aktech: Ah yes, makes sense

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
aktech: By footnote you mean at the bottom of the post?

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
```diff
@@ -0,0 +1,520 @@
...
+
+This was run on AMD Instinct MI50 32GB with 128 CPU cores.
+![NumPy cs CuPy AMD](/images/2022/02/numpy_vs_cupy_amd.png)
+
+
+#### On NVIDIA GPU:
+
+![NumPy cs CuPy NVIDIA](/images/2022/02/numpy_vs_cupy_nvidia.png)
+
+This was run on a NVIDIA TITAN RTX with 48 CPU cores. The GPU vs CPU plot is what you would
```
rgommers, line 457: Qgpu3 should have a Threadripper 3970X CPU, which is 32 physical (64 virtual) cores. Is 48 right here @aktech, @IvanYashchuk?

IvanYashchuk: NVIDIA TITAN RTX is on Qgpu2 and it has Threadripper 2970WX 24-Core (48 virtual) CPU. I don't know what machine was used for the plots.

aktech: I used QGPU2 for this, hence 48 cores (counting virtual cores)

rgommers: Ah okay, thanks. Let's say "24 physical cores" then? I think that's the normal way to talk about N-core CPUs.

rgommers: I'd going to assume it's the same for the other CPU, and update that to "64 physical cores".

posts/2022/02/making-gpus-accessible-to-pydata-ecosystem-via-array-api.md
rgommers: Yes, but I changed my mind on that one - just adding a few details in brackets.

rgommers (approved):
Alright, this LGTM now, in it goes. Thanks @aktech! And also @IvanYashchuk!

PR #331, opened 2022-06-22
adding the blog migration script to the new website

noatamir:
As suggested by @gabalafou I've added my blog migration script here for and I will also share a zip of the output via slack so anyone can open PRs. There are a bunch(!) to do now but it is substantially less work per post (though it does still take some time :wink:  

Here is a description of the process I had with the last few posts:

0. create a branch blog-filename
1. Add the md file under posts/year/month
2. Add a description in the header (usually easy to do based on an intro line in the first paragraph)
3.  Images are added using a folder structure similar to what we had before, but instead of the images folder we have a public folder and each post has its own unique folder. The pattern is something like this: public/posts/year/month/blog-file-name-as-folder-name/hero.svg
    3.1 Add the images if they exist, check for correct sizes (Hero, 1440x696. feature 391x174), and alt text. The easiest way to check if the blog had an image is to open it and see.  
    3.2 Add generic images with their alt text (we have a few already and I can go later and create more variations)
4. Check, using search, if there are any links to our github repository which need to be updated (e.g. search for "github.com/Quansight-Labs"). If you need to add files put them in them public folder similar to the images.
5. have a look around that everything look alright in the header and text. maybe check that images have alt-text if you have time. When ready commit all changes.
6. Open a PR & check the file blog_links.md (search for the file name) to see if the blog has links to other blogs, labs or llc. Please add the following to the commit message, and if there were links found paste them in so we can keep track and open issues to update this blog when the other one is migrated as well:

```
## Links to other Quansight blogs:

no labs links found
no llc links found
```

Any chance we can get a few more hands involved? It would go a lot faster and I can just share this via Notion with a zip of all the generated files so no one needs to worry about the migration script really.

### DIFF
--- a/migration/blogs_migration.py
+++ b/migration/blogs_migration.py
@@ -1,0 +1,176 @@
+from datetime import datetime
+import re
+import os
+from pathlib import Path
+from fnmatch import fnmatch
+from pandas import read_csv
+
+# Inputs:
+#          CSV from notion DB for new labels and author slugs
+#          download the zip file of the old website repo and unzip it in your Downloads folder
+#               (or clone it and update the root variable to the correct location)
+#          Have a ~/Documents/Quansight/New_Website/ directory where all the new files will be generated
+#          this includes the new md files following the same structure, and at the root of this dir a blog_links.md
+#          file which will have the list of all files created and detail if they had links to other blogs, labs or llc
+#          for manual editing later on.
+#          for questions contact @noatamir on gh or slack, or simply ask her for the output :wink:
+
+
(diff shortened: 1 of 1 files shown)


rgommers:
> Any chance we can get a few more hands involved? It would go a lot faster and I can just share this via Notion with a zip of all the generated files so no one needs to worry about the migration script really.

We have some people with available bandwidth I think, so this should be possible. Let's discuss on Slack, including @trallard.

rgommers:
Merging this before archiving the repo.

PR #333, opened 2022-07-07
Fix 404 pages

gabalafou:
Before this PR, the 404 page leaks the translation machinery, so you get a page with code in the h1 and div:

After this PR, you get a normal looking 404 that says "Not Found":


trallard:
LGTM 🚀

